PR #71, opened 2019-03-14
Hrosenbe/weat 96 update datamanager container

hrosenbe:
This pull requests changes all of the scripts in the auctionDataManager Docker container image to use Cassandra rather than MongoDB.
Additional changes:
 * Removed unneeded newDbLoader
 * Remove shard/replicas tests from dbPrep and dbLoader as these are mongodb concepts


hrosenbe:
Yes, the image files are only used by the dbLoader, which now always runs in the auctiondatamanager container.  The images are contained in the auctiondatamanager container image.

PR #72, opened 2019-03-14
use tie to output stderr to logs

bhoflich:


hrosenbe:
I verified that this works for information printed directly to STDERR (e.g. print STDERR "test error\n";).  

However it does not work for capturing output that goes to STDERR from a command run with system() or backticks ``, which is where most of the error output comes from in the run harness.  It looks like capturing that output will require more significant changes, such as wrapping all backtick commands using [Capture::Tiny](https://metacpan.org/pod/Capture::Tiny). Because that is a significant effort, I think that we should merge this code and then handle that change as a separate issue.

PR #82, opened 2019-04-09
Bhoflich/weat 125 fix seqnum in weathervane results

bhoflich:
A collection of smallish partially related updates.

Note: RunOnlyRunProcedure is currently broken following the hierarchical log changes.
I opened a separate task on this.
The changes to that specific file in this PR were for completeness but will need to be updated in a future PR.

### DIFF
--- a/runHarness/RunProcedures/PrepareOnlyRunProcedure.pm
+++ b/runHarness/RunProcedures/PrepareOnlyRunProcedure.pm
@@ -90,7 +90,7 @@ sub run {
 		print SEQFILE $nextSeqNum;
 		close SEQFILE;
 	}
-	my $seqnum = $majorSeqNum . "." . $minorSeqNum;
+	my $seqnum = $majorSeqNum . "-" . $minorSeqNum;
 	$self->seqnum($seqnum);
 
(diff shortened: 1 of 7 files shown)


runHarness/RunProcedures/PrepareOnlyRunProcedure.pm
```diff
@@ -52,36 +52,49 @@ override 'initialize' => sub {
...
 	else {
-		$seqnum = 0;
-		open SEQFILE, ">$sequenceNumberFile";
+		$minorSeqNum = 0;
+		open SEQFILE, ">$minorSequenceNumberFile";
 		my $nextSeqNum = 1;
 		print SEQFILE $nextSeqNum;
 		close SEQFILE;
 	}
+	my $seqnum = $majorSeqNum . "." . $minorSeqNum;
```
hrosenbe: I approved this but on further testing I would like to request one change.  Can you concatenate the major and minor numbers with a "-" rather than a ".".  Otherwise spreadsheet programs will treat the result as a floating point number rather than a string and may truncate.  E.g. run "42.0" displays as "42", with the 0 not displayed.

PR #84, opened 2019-04-16
Hrosenbe/weat 106 remove automatic hostname

hrosenbe:
This pull request represents a significant change in how Weathervane operates.  Instead of generating hostnames for services using the previous convention-based configuration, Weathervane now requires users to explicitly declare dockerHosts and kubernetesClusters and then provide a mapping from appInstances or services to those hosts or clusters.

This pull request also includes a few small changes:
- Updated version number to 2.0.0
- Removed targetUtilization run strategy
- default numWorkloads to 1
- Allow using either numXxServers or xxServers but not both


weathervane.pl
```diff
@@ -44,126 +43,134 @@ package Weathervane;
...
-		$weathervane_logger->debug( "For host ", $hostParamHashRef->{'hostName'}, " the Param hash ref is:" );
-		my $tmp = $json->encode($hostParamHashRef);
-		$weathervane_logger->debug($tmp);
-		$retArrayRef = createHost( $hostParamHashRef, $runProcedure, $instance, $ipToHostHashRef );
-
-		my ($createdNew, $host) = @$retArrayRef;
+	# If the xxxServerHosts was defined for this serviceType, then use 
+	# the right host from that list as determined by the instanceNum.
+	# The assignment of host to instance wraps if the instanceNum is greater
+	# than the number of hot names specified.
```
bhoflich: typo: hot->host

weathervane.config.docker
```diff
@@ -0,0 +1,60 @@
+{
+  # This is a temporary version of a docker config file.
+  # We can't use the fixed config sizes because the auctionBidServer does
+  # run in Docker-only yet.
```
bhoflich: typo?: "does run in Docker-only yet" -> "does not run in Docker yet" 
Here and below.

weathervane.config.docker
```diff
@@ -0,0 +1,60 @@
...
+    { "name" : "AuctionDriver5" }, { "name" : "AuctionDriver6" }, 
+    { "name" : "AuctionDriver7" }, { "name" : "AuctionDriver8" }, 
+    { "name" : "AuctionApp1", }, { "name" : "AuctionApp2", }, { "name" : "AuctionApp3", }, 
+    { "name" : "AuctionBid1", }, 
+    { "name" : "AuctionWeb1", }, 
+    { "name" : "AuctionCs1", }, { "name" : "AuctionCs2", }, { "name" : "AuctionCs3", }, 
+    { "name" : "AuctionMsg1", }, 
+    { "name" : "AuctionDb1", }, 
+    { "name" : "AuctionNosql1", }, 
+  ],
```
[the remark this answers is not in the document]
hrosenbe: The reason that this can't be a simple list of Strings is that this is where a user would specify other attributes of a dockerHost.  For example:

"dockerHosts" : [  { "name" : "dockerHost1", "dockerPort" : 2375, "dockerNet" : "host", } ],

We need to be able to specify those attributes on a per-host basis

weathervane.config.docker.hierarchical
```diff
@@ -0,0 +1,53 @@
...
+      "configurationSize": "custom",
+      "webServers" : [ {} ],
+      "webServerCpus" : "1",
+  	  "appServers" : [ {}, {}, {}, ],
+      "appServerCpus" : "1",
+  	  "auctionBidServers" : [],
+      "auctionBidServerCpus" : "1",
+  	  "msgServers" : [ {} ],
+      "msgServerCpus" : "1",
+      "coordinationServers" : [ {} ],
```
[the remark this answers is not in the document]
hrosenbe: Well, this is just an example and it needs will need to be updated once the bidService is running in a docker configuration.  But you are right that until the single zookeeper config is working this config won't work.

weathervane.config.docker
hrosenbe: Thanks.  Fixing in a new commit

weathervane.pl
hrosenbe: fixing in a new commit

PR #85, opened 2019-04-17
Restrict config directories relative to current working directory

bhoflich:
Set weathervaneHome to the current working directory
Adjust use lib for 'runHarness' to be relative to weathervane.pl
Use Parameters.pm default for $tmpDir and make it a constant
Move all relative path adjustments to weathervane.pl
Move logger setup to top of weathervane.pl script
Move help usage code up in weathervane.pl to be more usable
Print run number also with prepareOnly

Signed-off-by: Benjamin Hoflich <45051453+bhoflich@users.noreply.github.com>

notes:
There is still usage of /tmp.
Only $tmpDir is a 'fixed' location, but all other config directories are now relative to cwd.

### DIFF
--- a/runHarness/Parameters.pm
+++ b/runHarness/Parameters.pm
@@ -38,6 +38,8 @@ my @nonInstanceListParameters = ('userLoadPath', 'kubernetesClusters', 'dockerHo
 								 'msgServerHosts', 'coordinationServerHosts', 'dbServerHosts', 'nosqlServerHosts');
 my @runLengthParams = ( 'steadyState', 'rampUp', 'rampDown'  );
 
+my @filterConstants = ( 'tmpDir' );
+
 sub getParamDefault {
 	my ($key) = @_;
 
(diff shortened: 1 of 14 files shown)


runHarness/Parameters.pm
```diff
@@ -2420,119 +2429,109 @@ $parameters{"msgServerSuffix"} = {
...
+"This is the directory under which Weathervane stores configuration files for the various services relative to the current working directory.",
 	"showUsage" => 0,
 };
 $parameters{"dbLoaderImageDir"} = {
 	"type"    => "=s",
 	"default" => "images",
 	"parent"  => "dataManager",
 	"usageText" =>
-"This is the directory in which the images used by the dbLoader are stored.\n\tIt must either include the full path to the directory\n\tor be relative to weathervaneHome.",
+"This is the directory in which the images used by the dbLoader are stored relative to the current working directory.",
```
hrosenbe: This parameter can actually be removed since the images are now stored in the data manager's docker image

runHarness/Parameters.pm
```diff
@@ -2420,119 +2429,109 @@ $parameters{"msgServerSuffix"} = {
...
 $parameters{"distDir"} = {
 	"type"    => "=s",
 	"default" => "dist",
 	"parent"  => "runProc",
 	"usageText" =>
-"This is the directory for the executables and other artifacts needed to run Weathervane.\n\tIt must either include the full path to the directory\n\tor be relative to weathervaneHome.",
+"This is the directory for the executables and other artifacts needed to run Weathervane relative to the current working directory.",
 	"showUsage" => 0,
 };
 $parameters{"dbScriptDir"} = {
```
hrosenbe: The dbScriptDir parameter and all other references to dbScriptDir can be eliminated because the dbScripts are not located in the postgresql docker image.

PR #95, opened 2019-05-10
Hrosenbe/weat 134 simplify host specification

hrosenbe:
This pull request has two changes:
1. Add an appInstanceCluster parameter to specify the kubernetes cluster on which to run an appInstance.  The exisiting parameter appInstanceHost must now refer to a dockerHost.
1. The user can now specify the name of docker hosts in appInstanceHost, webServerHost, etc. that were previously not defined in a dockerHosts parameter.  dockerHosts specified this way use the defaults for the other parameters that apply to dockerHosts: dockerPort, dockerNet, vicHost.

This pull also updates the example config files to reflect this change, and has a minor fix for checking the format of the cpu and mem parameters.


weathervane.config.docker.custom
```diff
@@ -0,0 +1,119 @@
...
+    { "name" : "AuctionDriver1", "dockerPort" : 2376, "dockerNet" : "host" }, 
+    { "name" : "AuctionDriver1", "dockerPort" : 2376, "dockerNet" : "host" }, 
+    { "name" : "AuctionDriver1", "dockerPort" : 2376, "dockerNet" : "host" }, 
+    { "name" : "AuctionDriver1", "dockerPort" : 2376, "dockerNet" : "host" }, 
+    { "name" : "AuctionDriver1", "dockerPort" : 2376, "dockerNet" : "host" }, 
+    { "name" : "AuctionDriver1", "dockerPort" : 2376, "dockerNet" : "host" }, 
+    { "name" : "AuctionWeb1", "dockerPort" : 2376, "dockerNet" : "bridge" }, 
+    { "name" : "AuctionBid1", "dockerPort" : 2375, "dockerNet" : "bridge" }, 
+    { "name" : "AuctionBid2", "dockerPort" : 2376, "dockerNet" : "host" }, 
+    { "name" : "AuctionBid3", "dockerPort" : 2376, "dockerNet" : "host" },
```
bhoflich: 1 "bridge" and 2 "host" for AuctionBids?  Is that intended?

weathervane.config.docker.micro.oneHost
```diff
@@ -0,0 +1,17 @@
+{
+  "dockerNamespace" : "yourRepository",
+  
+  "numDrivers": 8,
```
bhoflich: micro.oneHost probably does not need 8 drivers?

weathervane.config.docker.custom
hrosenbe: Yes.  This config file is just intended to show the variety of things you can specify.  I wouldn't expect anyone to use it this way.

weathervane.config.docker.micro.oneHost
hrosenbe: Good catch.  I fixed this.

PR #97, opened 2019-05-19
Put total_memory in rabbitq.conf

hrosenbe:
There was a bug in the entrypoint script for the RabbitMQ image that was preventing some of the configuration parameter being placed into the rabbitmq.conf file.  In addition, the RabbitmqDockerService wasn't setting the RABBITMQ_MEMORY environment variable, and so when Docker memory was limited the process was being OOM killed.

This pull request fixes both issues.


runHarness/Services/RabbitmqDockerService.pm
```diff
@@ -61,12 +61,24 @@ override 'create' => sub {
...
+	my $totalMemory = $1;
+	my $totalMemoryUnit = $2;
+	if (lc($totalMemoryUnit) eq "gi") {
+		$totalMemoryUnit = "GB";
+	} elsif (lc($totalMemoryUnit) eq "mi") {
+		$totalMemoryUnit = "MB";
+	} elsif (lc($totalMemoryUnit) eq "ki") {
+		$totalMemoryUnit = "kB";
+	}
+	$envVarMap{"RABBITMQ_MEMORY"} = $self->internalPortMap->{$totalMemory$totalMemoryUnit};
```
bhoflich: Scalar found where operator expected at runHarness/Services/RabbitmqDockerService.pm line 80, near "$totalMemory$totalMemoryUnit"
        (Missing operator before $totalMemoryUnit?)

hrosenbe: This is just a bizarre typo and I'm not sure how I missed it.  I will fix this immediately

PR #104, opened 2019-06-05
Hrosenbe/weat 115 dockerize wkld driver add loadbalancer

hrosenbe:
This pull request covers a number of changes:

- The workloadDriver can now run on a Kubernetes cluster.
- Weathervane supports k8s clusters that use LoadBalancer services for ingress
- All of the kubernetes manifests have been updated to the latest version
- You can now specify the context for a cluster from a kubeconfig when defining a kubernetesCluster in a config file
- The runState returned from the workload driver now includes current status and more descriptive interval names.
- Some minor fixes and adjustments to the micro configuration

The most obvious change is the specification of Kubernetes clusters in the configuration file.  Kubernetes clusters are now specified like this:

"kubernetesClusters" : [ 
    { 
      "name" : "appCluster", 
      "kubeconfigFile" : "/root/.kube/config",
      "kubeconfigContext" : "cluster-context-1",
      "useLoadBalancer" : true,
    },
    { 
      "name" : "driverCluster", 
      "kubeconfigFile" : "/root/.kube/config",
      "kubeconfigContext" : "cluster-context-2",
      "useLoadBalancer" : true,
    },
  ],

The changes are:

- The parameter kubernetesConfigFile has changed to kubeconfigFile
- the parameter kubeconfigContext specifies the context from within the file.  If left blank it will use the current-context.
- If useLoadBalancer is set to true (the default), the nginx service and the wkldcontroller service will use a LoadBalancer service rather than a nodeport.


hrosenbe:
I forgot.  This pull request also includes changes so that the fixedConfigs are applied to the workloadDrivers as well.  fixedConfigs.json now includes the number of drivers and the driver tuning.  The number of drivers can be over-ridden in the config file.  The number of driver nodes actually run is numDrivers from the fixedConfig times the number of appInstances.

runHarness/ComputeResources/KubernetesCluster.pm
```diff
@@ -517,7 +691,45 @@ sub kubernetesGetLogs {
 
 	}
 	
-	return 0;
+	return 1;
```
bhoflich: The return value changed, but nobody checks it.
Does it even need a return value? (and for kubernetesCopyFromFirst)

configFiles/kubernetes/auctionworkloadcontroller.yaml
```diff
@@ -0,0 +1,93 @@
...
+  template:
+    metadata:
+      name: wkldcontroller
+      labels:
+        app: auction
+        tier: driver
+        type: controller
+    spec:
+      containers:
+      - image: hrosenbe/weathervane-auctionworkloaddriver:2.0.0
```
bhoflich: Maybe not related to this pull request but,
Should 'hrosenbe' be a replaced with a generic/invalid value in all of the yaml files?
After this release, those images will be constantly changing for whatever next releases.

hrosenbe: Good point.  I will change this now.

runHarness/AppInstance/AuctionKubernetesAppInstance.pm
```diff
@@ -153,18 +139,38 @@ override 'getEdgeAddrsRef' => sub {
...
+		while (!$ip) {
+			my $cmdFailed;
+		  	($cmdFailed, $ip) = $cluster->kubernetesGetLbIP("nginx", $self->namespace);
+		  	if ($cmdFailed)	{
+		  		$logger->error("Error getting IP for frontend loadbalancer: error = $cmdFailed");
+		  	}
+		  	if (!$ip) {
+		  		sleep 10;
+		  	}
+			$logger->debug("Called kubernetesGetLbIP: got $ip");
```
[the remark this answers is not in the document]
hrosenbe: Makes sense.  Fixed.

weathervane.pl
```diff
@@ -654,6 +680,34 @@ sub getComputeResourceForInstance {
...
+				if (!$workloadParamHashRef->{"numDrivers"} && ($#{$workloadParamHashRef->{"drivers"}} < 0)) {
+					my $numAppInstances = $workloadParamHashRef->{"numAppInstances"};
+					if (!$numAppInstances) {
+						my $instancesListRef = $workloadParamHashRef->{"appInstances"};
+						$numAppInstances = $#{$instancesListRef} + 1;					
+					}
+					$workloadParamHashRef->{$key} = $config->{$key} * $numAppInstances;
+				}
+			} else {
+				$workloadParamHashRef->{$key} = $config->{$key};
```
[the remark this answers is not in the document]
hrosenbe: I have created a Jira issue for this.  It isn't a straightforward change because we need to determine whether the value being overwritten is something the user set in the config file (or command-line) or is just the default.  We don't want to print a message when we override the default.

PR #109, opened 2019-06-12
Hrosenbe/weat 36 fix run strategies

hrosenbe:
This pull request includes the following changes:
- The fixed runStrategy uses a new QoS definition in which it must pass a certain number of independent QoS periods in order to pass.  
  - The number of QoS periods is defined by the numQosPeriods parameter and defaults to 3.
  - The length of a QoS Period is defined by the qosPeriodSec and defaults to 300.
  - The runLength and steadyState parameters have been removed.  The default for rampUp is now 300 and for rampDown is 120.
 - The findMaxMultiRun run procedure uses the new QoS on each run, but will end sub-runs early if any QoS period fails.
- Fixed a bug that caused the findMaxMultiRun runStrategy to always use the results of the first run as the decision criteria.
- This pull request also includes an experimental workloadProfile that is under development.  It should not yet be used.


runHarness/WorkloadDrivers/AuctionWorkloadDriver.pm
```diff
@@ -1144,33 +1154,12 @@ sub startRun {
 					my $runLengthMinutes = $totalTime / 60;
 					my $impl             = $self->getParamValue('workloadImpl');
 
-					if (!$usingFindMaxLoadPathType) {
+					if ($usingFixedLoadPathType) {
 						$console_logger->info(
 							"Running Workload $workloadNum: $impl.  Run will finish in approximately $runLengthMinutes minutes."
```
bhoflich: $runLengthMinutes should be rounded for this log message to prevent long strings of decimals when it is not divisible by 60.
Running Workload 1: auction.  Run will finish in approximately 6.33333333333333 minutes.

runHarness/Parameters.pm
```diff
@@ -938,39 +938,40 @@ $parameters{"description"} = {
...
-	"default" => "",
+	"default" => "300",
 	"parent"  => "runProc",
 	"usageText" =>
-"This is the length of the ramp-up period for the run.\n\tIt overrides the ramp-up length set by the runLength parameter.\n\tThe number of users is slowly increased over the course of the ramp-up period.",
+"This is the length of the ramp-up period for the run.",
 	"showUsage" => 1,
 };
-$parameters{"steadyState"} = {
+$parameters{"numQosPeriods"} = {
```
bhoflich: It is possible to specify 0 for numQosPeriods -- which although doesn't make sense, it kinda of runs but encounters uninitialized value issues.

runHarness/RunProcedures/RunProcedure.pm
```diff
@@ -226,7 +232,7 @@ sub followRunProgress {
 	my ($self, $tmpDir)         = @_;
 	my $console_logger = get_logger("Console");
 	my $rampUp         = $self->getParamValue('rampUp');
-	my $steadyState    = $self->getParamValue('steadyState');
+	my $steadyState = $self->getParamValue('qosPeriodSec');
```
bhoflich: For fixed, this causes start/stop stats on hosts to span qos periods (if there are more than 1 -- otherwise the stop happens during rampdown).

runHarness/WorkloadDrivers/AuctionWorkloadDriver.pm
hrosenbe: Good catch.  Will fix.

runHarness/RunProcedures/RunProcedure.pm
hrosenbe: That is ok.  We just want to get stats for a time while in steadyState.  I am only collecting for one QoS period because the run may exit after the first period, and detecting that would be a more significant change.  
I would prefer that stats collection didn't extend into rampdown, but it is ok.  The rampDown doesn't actually ramp the load down, it just continues to run at the same load for a few minutes after the last QoS period to ensure that shut-down effects don't mess up the QoS stats.  I would like to change the name of the interval from rampDown, but it would need to be changes in a lot of places.

runHarness/Parameters.pm
hrosenbe: Good point.  I will add a test for this.

runHarness/Parameters.pm
```diff
@@ -940,7 +940,7 @@ $parameters{"description"} = {
 
 $parameters{"rampUp"} = {
 	"type"    => "=i",
-	"default" => "300",
+	"default" => "420",
```
bhoflich: This is becoming a long time.  Is this ramp needed for every step or primarily just the first one?

hrosenbe: The problem is that in the micro configuration that shortened run of the appServerWarmer is leading to response-time problems in the initial portion of the first QoS period. For larger configurations we could probably have a shorter ramp-up.
We can experiment with these parameters when we do our sizing tests and see if there is a tradeoff between appServerWarmer time and rampUp time.  I agree that it would be better to have a shorter rampup, although given the start-up and shut-down time, a couple of minutes isn't a huge problem.

hrosenbe: Also, I'm not clear if this is what you were asking, but the rampUp is only before the first QoS period (at the start of the run).

bhoflich: My point was that for a findMaxMultiRun, rampup happens for every step (sub run).

PR #113, opened 2019-07-09
Hrosenbe/weat 192 workload driver fixes

hrosenbe:
This pull request contains fixes for a few issues in the workload driver:
- When running multiple appInstances, load paths that finish early now continue to run at their max passing load.
- The verifyMax phase of the findMaxLoadPath will now take larger steps if the levels keep failing, which will reduce the duration of runs that have a lucky pass on the first QoS period.
- Fix some null pointer exceptions that occured due to rare short stats intervals
- Don't retry channel allocation failures.  This retry could cause an infinite loop that burned CPU and heap.

This also contains some minor adjustments to the fixed configs.


hrosenbe:
That error was due to a cut-and-paste typo. I have fixed it.

PR #117, opened 2019-08-21
Hrosenbe/weat 203 lower after reload

hrosenbe:
This pull request covers a number of issues:
- Addresses problems that caused the performance to be lower after a db load
- Adjusts number and priority of threads in auctionApp and bid service to fix changing mix issues
- Adjust ramping down of load after failures
- Enable better tuning of dbLoader and dbPrep
- Tune cassandra compaction
- Fix issues with slow queries cause preparing data to faile


runHarness/ComputeResources/KubernetesCluster.pm
```diff
@@ -852,7 +852,7 @@ sub kubernetesTopPodAllNamespaces {
 				$logger->info("kubernetesTopPodAllNamespaces top pod failed: $cmdFailed");
 			}
 			my $time;
-			$cmd = "date +%H:%M";
+			$cmd = "date -u";
```
bhoflich: Should the date command in kubernetesTopNode be similarly changed?

runHarness/DataManagers/AuctionKubernetesDataManager.pm
```diff
@@ -248,31 +251,31 @@ sub pretouchData {
...
+	};
+	print $logHandle "Exec-ing /pretouchData.sh\n";
+	$logger->debug("Exec-ing /pretouchData.sh in cassandra-0");
+	my ($cmdFailed, $outString) = $cluster->kubernetesExecOne("cassandra", "/pretouchData.sh", $namespace);
+	if ($cmdFailed) {
+		$logger->debug( "Pretouching data failed for workload $workloadNum, appInstance $appInstanceNum. \$cmdFailed = $cmdFailed" );
+		print $logHandle "Pretouching data failed for workload $workloadNum, appInstance $appInstanceNum. \$cmdFailed = $cmdFailed\n";
+		return 0;
+	}
+	else {
```
bhoflich: Is this else block supposed to exist like this? logging "failed"?
Also, the code below is unreachable. (and it does not print $logHandle).

runHarness/RunProcedures/PrepareOnlyRunProcedure.pm
```diff
@@ -32,7 +32,7 @@ with Storage( 'format' => 'JSON', 'io' => 'File' );
 
 extends 'RunProcedure';
 
-has '+name' => ( default => 'Prepare-Only', );
+has '+name' => ( default => 'warmpare-Only', );
```
bhoflich: unintended change?

runHarness/Services/CassandraKubernetesService.pm
```diff
@@ -188,6 +188,26 @@ sub cleanData {
 	}	
 }
 
+sub waitForReady {
+	my ($self)   = @_;
+	my $logger = get_logger("Weathervane::Services::CassandraKubernetesService");
+	$logger->debug("waitForReady");
+	my $cluster = $self->host;
+	my $finished = 0;
+	while (!$finished) {
```
bhoflich: Should there be a guard against infinite loop with some type of timeout failure?

runHarness/ComputeResources/KubernetesCluster.pm
hrosenbe: Yes.  Good point.  I needed this change to simplify matching up timestamps in the logs with these times

runHarness/DataManagers/AuctionKubernetesDataManager.pm
hrosenbe: The log messages are a copy/past issue, but that is moot because the entire pretouchData path is not used.  This took a long time to execute and didn't improve performance.  I intended to take this all out and forgot.  I will do that now.

runHarness/RunProcedures/PrepareOnlyRunProcedure.pm
hrosenbe: Yup.  Must have had the cursor in the wrong window.

runHarness/Services/CassandraKubernetesService.pm
hrosenbe: Yes.  I will add this, although for now I am going to make it a long timeout because the compaction may take a long time in the large configs.  I will add an issue to track adjusting the timeout later.

PR #120, opened 2019-09-03
rework start/prepare/prepareDataServices to exit cleanly on error

bhoflich:
change kubernetesGetLogs error path to return 0 instead of exit
add a copy for wkldriver gc log

Signed-off-by: Benjamin Hoflich <45051453+bhoflich@users.noreply.github.com>


runHarness/AppInstance/AppInstance.pm
```diff
@@ -707,15 +711,23 @@ sub isRunningAndUpServices {
 	if ( !$allIsRunning ) {
 		$console_logger->error(
 			"Couldn't bring to running all $serviceTier services for appInstance $appInstanceNum of workload $workloadNum." );
-		exit 1;
+		if ($forked) {
+			exit;
```
hrosenbe: With the current implementation of callBooleanMethod..., the use of exit rather than exit 1 is correct.  However we will need to change these to exit 1 once we have fixed callBoolean... However, I do not want to hold this pull request up for those changes.

runHarness/ComputeResources/KubernetesCluster.pm
```diff
@@ -866,8 +866,8 @@ sub kubernetesGetLogs {
 	$logger->debug("Output: $outString");
 	my @names = split /\s+/, $outString;
 	if ($#names < 0) {
-		$console_logger->error("kubernetesGetLogs: There are no pods with label $serviceTypeImpl in namespace $namespace");
-		exit(-1);
+		$logger->debug("kubernetesGetLogs: There are no pods with label $serviceTypeImpl in namespace $namespace");
+		return 0; #this can happen in the cleanupAfterFailure path
```
hrosenbe: Good catch.  I have seen this message and it did need to be removed.

PR #123, opened 2019-09-06
Hrosenbe/weat 185 medium sizing

hrosenbe:
Changes in this pull request:
- Always assign a random uid to the rabbitmq user in the rabbitmq pod.  This prevents rabbitmq pods from sharing resource limits on the worker nodes.
- Remove the liveness and readiness probes from the rabbitmq pod.  These are not needed because the run harness and entrypoint script have the same functionality, and the probes were adding CPU overhead.
- Minor adjustments to the mix percentages that should have been made after some previous bug fixes stabilized the mix.
- Removed the readiness probes from the workload driver.  The run harness already has the same functionality, and have the probes on the pod greatly slowed the workload driver startup (since the next pod in a stateful set cannot start until the previous is ready).


dockerImages/rabbitmq/entrypoint.sh
```diff
@@ -19,6 +19,12 @@ trap 'sigusr1' USR1
 echo "Add weathervane domain to resolv.conf" 
 perl /updateResolveConf.pl
 
+newuid=$((4096 + RANDOM))
+olduid=$(id -u rabbitmq)
+echo "Assigning random uid $newuid to rabbitmq user with uid $olduid"
+usermod -u $newuid rabbitmq
+find / -user $olduid  -exec chown -h $newuid {} \;
+
```
bhoflich: Is it possible for this to conflict with an existing user id on the host (running docker)?

hrosenbe: That is why I am am adding 4096 to the random number.  The pre-existing UIDs are either low numbers (less than 1024) or high numbers (nobody == 65534).  Since RANDOM return random numbers from 1 to 32767, the uids that I am generating will go from 4097 to 36863.  There could still be a conflict with another randomly generated uid, but the impact of that should be minimal.

PR #126, opened 2019-09-16
Hrosenbe/weat 221 load stats sync

hrosenbe:
This pull request contains the following changes:
- Fixed an issue that caused a loadPath's load interval and stats intervals to get out of sync at high load.
- Increased casandra read timeout to allow for slow startup when running many appInstances
- Adjust CPU requests for small config to use less than 6 CPUs total for the sut.
- Remove cpu/memory requests for the appServer warmer
- limit the number of backend connections that will be used by nginx to any single backend.  This change helps avoid running into port range limits.
- When running with a NodePort (useLoadbalancer==false), only balance load across nodes labeled wvrole=sut, if any.  If no nodes are labeled wvrole=sut, then balance across all nodes except master nodes.


workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadPath/FindMaxLoadPath.java
```diff
@@ -232,6 +229,11 @@ private UniformLoadInterval getNextInitialRampInterval() {
 		curStatusInterval.setStartUsers(curUsers);
 		curStatusInterval.setEndUsers(curUsers);
 		curStatusInterval.setDuration(nextIntervalDuration);
+		
+		statsIntervalComplete = true;
+		curStatsInterval.setName(nextInterval.getName());
```
bhoflich: this line and -->

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadPath/FindMaxLoadPath.java
```diff
@@ -232,6 +229,11 @@ private UniformLoadInterval getNextInitialRampInterval() {
...
+		curStatsInterval.setUsers(curUsers);
+		curStatsInterval.setDuration(nextIntervalDuration);
```
bhoflich: this line were just set above. do they need to be set again?

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadPath/FindMaxLoadPath.java
```diff
@@ -433,6 +435,11 @@ private UniformLoadInterval nextInterval() {
 				curStatusInterval.setEndUsers(curUsers);
 				curStatusInterval.setDuration(warmupIntervalSec);
 				
+				statsIntervalComplete = true;
+				curStatsInterval.setName(nextInterval.getName());
+				curStatsInterval.setUsers(curUsers);
+				curStatsInterval.setDuration(warmupIntervalSec);
```
bhoflich: name and duration set again, as above.

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadPath/FindMaxLoadPath.java
```diff
@@ -454,7 +461,12 @@ private UniformLoadInterval nextInterval() {
 				curStatusInterval.setStartUsers(curUsers);
 				curStatusInterval.setEndUsers(curUsers);
 				curStatusInterval.setDuration(getQosPeriodSec());
-				
+
+				statsIntervalComplete = true;
+				curStatsInterval.setName(nextInterval.getName());
+				curStatsInterval.setUsers(curUsers);
+				curStatsInterval.setDuration(getQosPeriodSec());
```
bhoflich: name and duration set again.

runHarness/ComputeResources/KubernetesCluster.pm
```diff
@@ -537,8 +538,9 @@ sub kubernetesGetNodeIPs {
 	  $contextString = "--context=$context";	
 	}
 
+	# First try to get all nodes with label wvrole=sut
 	my $cmd;
-	$cmd = "kubectl get node --kubeconfig=$kubeconfigFile $contextString -o=jsonpath='{.items[*].status.addresses[?(@.type == \"ExternalIP\")].address}'";
+	$cmd = "kubectl get node --kubeconfig=$kubeconfigFile $contextString --selector=wvrole=sut -o=jsonpath='{.items[*].status.addresses[?(@.type == \"ExternalIP\")].address}'";
```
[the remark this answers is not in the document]
hrosenbe: The command will not fail is there are no labeled nodes, it will just return an empty string.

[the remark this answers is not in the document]
hrosenbe: The command will not fail if there are no labelled nodes.  It will just return an empty string.

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadPath/FindMaxLoadPath.java
hrosenbe: The ones above are setting the curStatusInterval, while these are setting the curStatsInterval. These  are used for different purposes.  It is possible that we don't need both now that I am handling the statsIntervals this way, but I don't want to break something trying to figure it out. At some point I would like to clean up the loadPath code, but not now.

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadPath/FindMaxLoadPath.java
hrosenbe: Same thing, curStatusInterval Vs curStatsInterval

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadPath/FindMaxLoadPath.java
hrosenbe: Same thing, curStatusInterval Vs curStatsInterval

PR #127, opened 2019-09-17
Hrosenbe/weat 225 limit parameter

hrosenbe:
This pull request adds two new parameters:
- useAppServerLimits: When set to true, the appServer pods will use appServerCpus and appServerMem as both requests and limits in the Kubernetes manifest.  Otherwise they will only be used as requests.
  - The default for this parameter is currently true, but that may change.
- useKubernetesLimits: When set to true, all services of the auction app will use their cpu and mem values as both requests and limits in the Kubernetes manifests.  Otherwise they will be used only as requests.
  - The default for this parameter is false.


runHarness/Services/TomcatKubernetesService.pm
```diff
@@ -112,11 +112,21 @@ sub configure {
...
-				}
-				elsif ( $inline =~ /\s\s\s\s\s\s\s\s\s\s\s\smemory:/ ) {
-					print FILEOUT "            memory: " . $self->getParamValue('appServerMem') . "\n";
+				if ( $inline =~ /(\s+)resources/ )  {
+					my $indent = $1;
+					print FILEOUT $inline;
+					print FILEOUT "$indent  requests:\n";
+					print FILEOUT "$indent    cpu: " . $self->getParamValue('appServerCpus') . "\n";
+					print FILEOUT "$indent    memory: " . $self->getParamValue('appServerMem') . "\n";
+					if ($self->getParamValue('useAppServerLimits') || $self->getParamValue('useKubernetesLimits')) {
```
bhoflich: In the rare edge case where these evaluate to false || true, which doesn't make much sense, should an error be thrown?

hrosenbe: My thinking is that we may change the default for useAppServerLimits to false, but there may be a case where we (or someone) wants to experiments with the effects of limits on all services of an instance.  In this case false || true is ok.  The useAppServerLimits is really a special case that allows us to tune a fixedConfig to a target appServer CPU usage.

An example of when someone might want to use limits on all services would be to verify that the limits ensure that an appInstance remains performant even in the case of an overloaded system.  You could have one appInstance with limits, and many others without.  In this case, the instance with limits should always get its resources, while the other may not, or may even be killed.

PR #129, opened 2019-09-23
Hrosenbe/1qos no limit

hrosenbe:
- Require only one passing QoS interval by default.  Requiring 3 did not improve repeatability or greatly affect results
- By default, do not use CPU/Memory limits for the appServer.  This enables maxing out the host CPU with fewer than the max number of app instances that fit on a cluster
- Increase maxUsers for micro and small to support running without limits on faster systems
- Revert nginx tuning change that I was testing but had intended to revert before committing
- Minor updates to example config files
- Save fixedConfigs file for each run

Following the initial creation of the pull request I added the following changes:
- Added an initialRampRateStep parameter and made this part of the fixedConfigs.  So now we can specify the rate step for each config size.
- Fixed a bug in the AuctionBidService that allowed null pointer exceptions
- Turned on debug logging for FinxMaxLoadPath.  These messages should help us debug the sporadic hangs that we have seen.  We will turn these off before release.  I also updated the log messages to make it easier to distinguish among appInstances.  The log messages show up in run_W1.log.


workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadPath/FindMaxLoadPath.java
```diff
@@ -131,7 +131,7 @@ public void initialize(String runName, String workloadName, Workload workload, L
 		logger.debug("initialize " + this.getName() + ": minUsers = {}, maxUsers = {}", getMinUsers(), maxUsers);
 
 		this.maxUsers = workload.getMaxUsers();
-		this.initialRampRateStep = maxUsers / 20;
+		this.initialRampRateStep = maxUsers / 40;
```
bhoflich: Does this now set the initialRampRateStep to 750 on small?

hrosenbe: Fixed by adding initialRampRateStep parameter

PR #131, opened 2019-10-11
Hrosenbe/weat 235 connect timeout

hrosenbe:
- The default name for the storageClass used by the various services is now weathervaneSC to help avoid accidentally using an existing class.
- Use two web servers for small to avoid resource limit with only one SUT worker node
- Reduce the CPu requests for all but the appServer to allow fitting more instance in one cluster
- Turn on limits on the appServer by default


runHarness/Parameters.pm
```diff
@@ -2352,7 +2352,7 @@ $parameters{"nginxUseNamedVolumes"} = {
 
 $parameters{"nginxCacheStorageClass"} = {
 	"type"      => "=s",
-	"default"   => "fast",
+	"default"   => "weathervaneSC",
```
bhoflich: Should the weathervane.config* files be similarly updated?

hrosenbe: Yes.  Done.

PR #137, opened 2019-10-30
Rebeccayo/weat 207 run pass fail

RebeccaYo:
This is the WEAT-207 bug **Replace zeros in weathervaneResults.csv**

When findmax fails (this does not typically occur and happens only when a maximum number of users is unable to be found), the use of the summary values caused 'uninitialized' warnings, which I disabled. Please let me know if this is not an acceptable solution.

Changelog:
- Add new column "Pass" to the left of the WvUsers column.
- When all AIs pass, Pass=Y. Otherwise, Pass=N.
- For findmax runs, if Pass=N, summary fields contain "NA"
- For non-findmax runs, and passing findmax runs, summary fields contain sum or average across all AIs
- Remove the column labeled "workload"
- Remove the column labeled "workloadProfile".
- Rename the column labeled "users-total" to WvUsers.


runHarness/WorkloadDrivers/AuctionWorkloadDriver.pm
```diff
@@ -1928,26 +1928,45 @@ sub getWorkloadStatsSummary {
 	my $overallAvgRT = 0;
 
 	my $numPassingAi = 0;
+	my $numTotalAi = 0;
 	my $appInstancesRef = $self->workload->appInstancesRef;
+	$numTotalAi = scalar @$appInstancesRef;
```
bhoflich: I think all other parts of the code used $# with +1 to get number of elements, although scalar should work also.

runHarness/WorkloadDrivers/AuctionWorkloadDriver.pm
```diff
@@ -1928,26 +1928,45 @@ sub getWorkloadStatsSummary {
...
+			
+	if ($numPassingAi == $numTotalAi) { # If all AIs pass, the run passes
+		$csvRef->{"Pass"}				= "Y";
 	} else {
-		$csvRef->{"overallAvgRT-total"} = 0;		
+		$csvRef->{"Pass"}				= "N";
 	}
+
+	# If this is a failed findMax run, print "NA" for summary fields
+	if ( ($csvRef->{"Pass"} eq "N") && ($self->workload->appInstancesRef->[0]->getParamValue('loadPathType') eq "findmax") ) {
```
bhoflich: Should it check for syncedfindmax also?

hrosenbe: Yes.  You need to check for syncedfindmax here as well.  I don't think that loadPathType existed in the code you were using originally.

RebeccaYo: Thank you for the catch. This has been fixed in commit ebc4db1.

PR #141, opened 2019-11-18
Hrosenbe/hint

hrosenbe:
This commit request has the following changes:
- Added a parameter called maxPassHint.  If this is set to a non-zero value, then the findMaxSingleRun run strategy will do its initial ramp up to that number of users, and then begin the findmax portion of the run.  The findmax will start with a step of 1/100 of the number of users, rather than the 1/10 that is used when maxPassHint is not provided.

Providing a starting hint will often, but not always make the run complete more quickly.


hrosenbe:
One additional point.  The maxPassHint is currently per-instance.  I am still undecided whether it should be the total passing users expected.

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadControl/loadPath/SyncedFindMaxLoadPath.java
```diff
@@ -218,11 +219,18 @@ private IntervalCompleteResult initialRampIntervalComplete() {
...
 				getIntervalStatsSummaries().add(rollup);
 			}
-		} 
+		} 		
+		if (((getMaxPassHint() > 0) && (curUsers == getMaxPassHint())) || 
+				((getMaxPassHint() == 0) && !prevIntervalPassed) || 
+				((curUsers + getInitialRampRateStep()) > maxUsers)) {
+			result.setPassed(false);
+		} else {
+			result.setPassed(true);
```
bhoflich: Should this be result.setPassed(prevIntervalPassed)?  Is it possible prevIntervalPassed can be false in this path (from line 217/218)?

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadControl/loadPath/SyncedFindMaxLoadPath.java
```diff
@@ -234,15 +242,18 @@ private UniformLoadInterval getNextInitialRampInterval(boolean prevIntervalPasse
 		logger.debug("getNextInitialRampInterval " + this.getName() + ": interval " + intervalNum);
 
 		/*
-		 * If we have reached max users, or the previous interval failed, then to go to
+		 * If the previous interval failed, then to go to
 		 * the FINDFIRSTMAX phase
 		 */
-		if (!prevIntervalPassed || ((curUsers + getInitialRampRateStep()) > maxUsers)) {
```
bhoflich: Is this check no longer needed in the non hint case also?

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadControl/loadPath/SyncedFindMaxLoadPath.java
hrosenbe: No.  There are only the three cases in which we should set passed to false and thus move to findFirstMax.  In particular we want to set passed to true even if prevIntervalPassed is false when maxPassHint is non-zero.  When running with the hint, the initialRamp always ramps up to the value of maxPassHint, even if one of the intervals along the way fails response-time.

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/loadControl/loadPath/SyncedFindMaxLoadPath.java
hrosenbe: This check was moved to the if clause that you commented on in initialRampIntervalComplete(), so it isn't needed here.

runmany.sh
```diff
@@ -1,22 +1,35 @@
...
 
-# The following invocation can be used to run the workload in maximum finding mode.
-./weathervane.pl --users=200 --runStrategy=findMax --initialRateStep=100 --minRateStep=50 --runLength=medium --description="initial test"  
+# The following invocation can be used to run the findMaxSingleRun runStrategy, 
+# even if the configuration file specifies the fixed runStrategy.
+#./runWeathervane.pl --configFile weathervane.config -- --runStrategy=findMaxSingleRun
 
+# The following invocation can be used to run the findMaxSingleRun runStrategy, 
+# with a different number of application instances.
+#./runWeathervane.pl --configFile weathervane.config -- --runStrategy=findMaxSingleRun --numApplicationInstances=2
```
bhoflich: Isn't it spelled:
--numAppInstances
?

hrosenbe: Yes.  Good catch.  I will fix

PR #147, opened 2019-12-19
building of executables takes place in a java 8 container

hrosenbe:
Changes in this pull request:
- When running buildDockerImages.pl, the java executables are now built using a java 8 docker container.  This eliminates the need for the user to have Java installed, and also avoids any issue with incompatible Java versions.  This also brings the behavior into line with the User's guide, which doesn't mention the need to have Java 8 installed to build the images.
- This pull request will trigger a 2.0.2 release.


buildDockerImages.pl
```diff
@@ -229,10 +230,22 @@ sub cleanupAfterBuild {
...
+my $version = `cat version.txt`;
+chomp($version);
+
 # Build the executables
 print "Building the executables.\n";
 print $fileout "Building the executables.\n";
-runAndLog($fileout, "./gradlew release");
+
+# Create a .gradle directory and map it into the container
+# This will speed susequent builds
```
RebeccaYo: Typo: Subsequent

buildDockerImages.pl
```diff
@@ -229,10 +230,22 @@ sub cleanupAfterBuild {
...
+my $cwd = getcwd();
+`mkdir -p $cwd/.gradle`;
+my $cmdString = "docker run --name weathervane-builder --rm "
+              . "-v $cwd/.gradle:/root/.gradle " 
+              . "-v $cwd:/root/weathervane -w /root/weathervane " 
+              . "--entrypoint /root/weathervane/gradlew openjdk:8 release";
```
RebeccaYo: I have a basic question here. The docker run command, that I can see, doesn't specify an image. Does it use a default image then?

RebeccaYo: I see, thank you.

PR #149, opened 2020-01-23
Reduce RabbitMQ Overhead

hrosenbe:
This pull request contains changes to reduce the overhead of RabbitMQ.  These are:
* Change the Liveness and Readiness probes to use curl to the management API rather than using _rabbitmqctl status_.  There is a known issue with rabbitmqctl causing high CPU usage when used in Kubernetes probes.
* Limit the number of erlang scheduler threads used by RabbitMQ to 1.  Otherwise RabbitMQ will launch as many scheduler threads as there are CPUs on the Kubernetes node.  This adds overhead, particularly on a heavily loaded system.  

These changes have been tested using Micro instances on 20 x 4vCPU worker nodes.  At lower numbers of instances there was no impact on peak WvUsers.  At higher numbers of instances I saw an improvement of 3 to 5%.


hrosenbe:
This pull request accidentally includes the changes from #150.  I am closing it and will recreate.

PR #150, opened 2020-01-23
Enable driver<->app communication with ClusterIP

hrosenbe:
* When running the workload drivers and applications on the same cluster, it is now possible to have the Nginx service created as a ClusterIP service.  In this case the communication between the drivers and the applications will only use cluster-internal IP addresses.
* The _useLoadBalancer_ parameter has been removed and replaced with the _appIngressMethod_ parameter.  The appIngressMethod parameter can take the values loadbalancer, nodeport, or clusterip.
    * The User's guide has been updated to reflect this change.
    * A check and message has been added to runWeathervane.pl to notify users for the change.
* In order to remove the need for cluster-external IP addresses for communication between the run harness and the workload controller, all such communication now uses _kubectl exec_ to run http commands inside the wkldcontroller container (using the httpie command).  
* Weathervane can now run on Minikube.  The documentation was been updated with Minikube specific information.
* As a workaround for clusters with DNS issues, the IP address of a Cassandra node is always used as the listen address, rather than the hostname.  In addition, on cassandra-0 it is added to the list of seeds.
* The resource requests have been removed from the AuctionDataManager pod so that it can be scheduled when running with small nodes.
* The wait time for all services starting has been increased to accommodate runs with large numbers of instances on slow nodes.
* Services are now started in batches of 10 with a 60 second inter-batch delay.  Previously all services were started simultaneously  This delay fixes issues with inability to start all instances on large runs.


doc/userDocs/usersGuide.md
```diff
@@ -1137,7 +1190,7 @@ Configuration parameters not mentioned here have an empty string as the default
 | `runStrategy`               | `findMaxSingleRun`            |
 | `numAppInstances`           | `1`                           |
 | `kubeconfigFile`            | `~/.kube/config`              |
-| `useLoadBalancer`           | `TRUE`                        |
+| `appIngressMethod`          | `loadbalancer                |
```
bhoflich: missing closing back quote?

dockerImages/runharness/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Copyright 2017-2019 VMware, Inc.
 # SPDX-License-Identifier: BSD-2-Clause
-FROM namespace/weathervane-centos7:version
+FROM hrosenbe/weathervane-centos7:intraCluster
```
bhoflich: Use namespace instead of hrosenbe?
What does intraCluster do?

runHarness/Utils.pm
```diff
@@ -255,6 +255,43 @@ sub callBooleanMethodOnObjectsParallel2 {
...
+sub callBooleanMethodOnObjectsParallel2BatchDelay {
+	my ( $method, $objectsRef, $param1, $param2, $batchSize, $delaySec ) = @_;
+	my $console_logger = get_logger("Console");
+	my @pids;
+	my $pid;
+	my $objectNum = 0;
+
+	foreach my $object (@$objectsRef) {
+		if ($objectNum == $batchSize) {
+			sleep($delaySec);
```
bhoflich: I suggest printing a log message (perhaps debug) when a sleep occurs.
Here in 2 param and in 3 param also.

configFiles/kubernetes/auctionDataManager.yaml
```diff
@@ -70,10 +70,6 @@ spec:
             - /loadData.pl
           initialDelaySeconds: 15
           periodSeconds: 10
-        resources:
-          requests:
-            cpu: 1
-            memory: 8Gi
```
bhoflich: Will this change impact db load time?

doc/userDocs/usersGuide.md
```diff
@@ -141,7 +140,32 @@ The image below shows the lines that need to be edited in the configuration file
...
+      clusters, you can choose either:
+        - `loadbalancer`:  Traffic from the workload drivers to the applications will
+          use the external IP address of a Kubernetes LoadBalancer service. Your cluster must 
+          support provisioning external IP addresses for LoadBalancer services. 
+        - `nodeport`:  Traffic from the workload drivers to the applications will
+          use the external IP addresses of the Kubernetes Nodes via a NodePort service. Your cluster must 
+          support exposing the node IP addresses externally using NodePort services.
+    - If the workload drivers and the applications are running on the same Kubernetes 
+      cluster, you can choose among:
+        - `none`: Traffic from the workload drivers to the applications will use a
```
bhoflich: Should 'none' be 'clusterip' ?

doc/userDocs/usersGuide.md
```diff
@@ -141,7 +140,32 @@ The image below shows the lines that need to be edited in the configuration file
...
+          use the external IP address of a Kubernetes LoadBalancer service. This 
+          may be desirable even with a single cluster in order to load-test the complete networking 
+          stack of your cluster. Your cluster must 
+          support provisioning external IP addresses for LoadBalancer services. 
+        - `nodeport`: Traffic from the workload drivers to the applications will
+          use the external IP addresses of the Kubernetes Nodes via a NodePort service. This 
+          may be desirable even with a single cluster in order to load-test the complete networking 
+          stack of your cluster. Your cluster must 
+          support exposing the node IP addresses externally using NodePort services.
+
```
bhoflich: There is a lot of duplication here.   Can the clusterip case and "desirable even with a single cluster" text be factored out to allow only 1 loadbalance and 1 nodeport section?

doc/userDocs/usersGuide.md
```diff
@@ -141,7 +140,32 @@ The image below shows the lines that need to be edited in the configuration file
 2. Update *kubeconfigFile* to the full path to your kubeconfig file.  The paths will be identical if using a single cluster for both the app and driver.
 3. Update *kubeconfigContext* parameters with the name of the context for your cluster from the kubeconfig file, such as `kubernetes-admin@kubernetes`.
     - Set an empty string (`""`) to use the current-context defined in the kubeconfigFile.
-4. Set *useLoadBalancer* to `true` if your cluster supports LoadBalancer services. Otherwise, set it to `false` to use a NodePort for ingress.
+4. Choose a value for *appIngressMethod* that reflects the method you want to use
+  for communication between the workload drivers and the applications.
+    - If the workload drivers and applications are running on different Kubernetes 
+      clusters, you can choose either:
+        - `loadbalancer`:  Traffic from the workload drivers to the applications will
+          use the external IP address of a Kubernetes LoadBalancer service. Your cluster must
```
bhoflich: 'Your cluster"  -- does this mean the driver cluster or the application cluster, or both?

doc/userDocs/usersGuide.md
```diff
@@ -141,7 +140,32 @@ The image below shows the lines that need to be edited in the configuration file
...
-4. Set *useLoadBalancer* to `true` if your cluster supports LoadBalancer services. Otherwise, set it to `false` to use a NodePort for ingress.
+4. Choose a value for *appIngressMethod* that reflects the method you want to use
+  for communication between the workload drivers and the applications.
+    - If the workload drivers and applications are running on different Kubernetes 
+      clusters, you can choose either:
+        - `loadbalancer`:  Traffic from the workload drivers to the applications will
+          use the external IP address of a Kubernetes LoadBalancer service. Your cluster must 
+          support provisioning external IP addresses for LoadBalancer services. 
+        - `nodeport`:  Traffic from the workload drivers to the applications will
+          use the external IP addresses of the Kubernetes Nodes via a NodePort service. Your cluster must
```
bhoflich: "Your cluster" again.

doc/userDocs/usersGuide.md
```diff
@@ -206,16 +230,39 @@ Handling clusters whose credential expire is discussed [below](#expire).
...
+The Weathervane workload driver interacts with the Weathervane application through a Kubernetes service fronting the application's web-server 
+tier. Weathervane 
+allows the user to select the type of Kubernetes service to be used.  Your selection will depend 
+on the types of services supported by your cluster and the goals of your tests.  
+
+Weathervane supports the use of 
+[LoadBalancer, NodePort, or ClusterIP services](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types).   You 
+indicate the type of service you want to use by setting the *appIngressMethod* parameter 
+to `loadbalancer`, `nodeport`, or `clusterip`.  The default value for this parameter 
+is `loadbalancer`.
```
bhoflich: Maybe add a note that for vanilla / custom created clusters (non cloud), that loadbalancer will likely not apply and this will need to be changed from the default value?

doc/userDocs/usersGuide.md
```diff
@@ -1356,4 +1409,79 @@ As a result, when using node labels it is recommended that you label all of your
...
+running.
+
+#### Weathervane Configuration
+
+In order to run Weathervane on Minikube, you need to use the following configuration 
+parameter values:
+
+| Parameter | Value |
+|-----------|-------|
+| appIngressMethod | clusterIp |
```
bhoflich: should be 'clusterip' ? (lower case i)

doc/userDocs/usersGuide.md
hrosenbe: Yes.  Will fix in next commit.

dockerImages/runharness/Dockerfile
hrosenbe: Good catch.  This is something that must have crept in because I aborted a build of the runharness container. intraCluster was what I was using for the version number (in version.txt).

runHarness/Utils.pm
hrosenbe: Added a debug log message for now.

configFiles/kubernetes/auctionDataManager.yaml
hrosenbe: No.  The requests are only used for admission control.  For micro the 1 CPU request is more than the CPU requests of the non-dataServices pods, and so I was running into situations where it couldn't start the data manager pod even though there would be enough space for all of the instances once the data was prepared.

doc/userDocs/usersGuide.md
hrosenbe: yes.  Good catch.  I initially had a slightly different interpretation of the appIngressMethod parameter, and was using _none_ rather than _clusterip_.

doc/userDocs/usersGuide.md
hrosenbe: Agreed.  I have refactored this to remove the duplication.

doc/userDocs/usersGuide.md
hrosenbe: The application cluster.  I will clarify

doc/userDocs/usersGuide.md
hrosenbe: will clarify

doc/userDocs/usersGuide.md
hrosenbe: agreed.  i will add this.

doc/userDocs/usersGuide.md
hrosenbe: yes.  I just pushed a commit with fixes for these comments.

doc/userDocs/usersGuide.md
hrosenbe: added

doc/userDocs/usersGuide.md
hrosenbe: done

doc/userDocs/usersGuide.md
```diff
@@ -141,29 +141,21 @@ The image below shows the lines that need to be edited in the configuration file
...
-        - `none`: Traffic from the workload drivers to the applications will use a 
-          cluster internal address provisioned by a Kubernetes ClusterIP service.  
-          This method will work on all clusters.
-        - `loadbalancer`: Traffic from the workload drivers to the applications will
+  for communication between the workload drivers and the applications. You can choose 
+  from among:
+    - `clusterip`: Traffic from the workload drivers to the applications will use a 
+          cluster internal address provisioned by a Kubernetes ClusterIP service.  This 
+          choice will only work if the workload drivers and applications are running 
+          on the same cluster.
```
bhoflich: I forgot to ask, what happens if you specify clusterip with multiple clusters?
Is there a check, or does it just run into errors, or?

hrosenbe: Good question.  There should be a check but there is not.  Currently it will just not work, and I don't know what errors it will give (as I only have a single cluster).  I will add a check.

doc/userDocs/usersGuide.md
```diff
@@ -241,7 +233,9 @@ Weathervane supports the use of
 [LoadBalancer, NodePort, or ClusterIP services](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types).   You 
 indicate the type of service you want to use by setting the *appIngressMethod* parameter 
 to `loadbalancer`, `nodeport`, or `clusterip`.  The default value for this parameter 
-is `loadbalancer`.
+is `loadbalancer`.  Node that cluster created manually using kubeadm will likely not 
+support LoadBalancer services.  In than case you will need to use either NodePort or
```
bhoflich: than -> that

runHarness/AppInstance/AuctionAppInstance.pm
```diff
@@ -111,21 +110,25 @@ override 'checkConfig' => sub {
...
 		return 0;
 	}
 
-	my $validAppInstanceSizes = $WeathervaneTypes::appInstanceSizes->{"auction"};
-	if ( $self->getParamValue('appInstanceSize') ~~ @$validAppInstanceSizes ) {
-		$console_logger->error("Workload $workloadNum, AppInstance $appInstanceNum: The AppInstance size for the Auction workload must be one of: @$validAppInstanceSizes");
-		return 0;
-	}
+    my $validAppInstanceSizes = $WeathervaneTypes::appInstanceSizes{"auction"};
+    if (!(grep { $self->getParamValue('configurationSize') eq $_ } @$validAppInstanceSizes)) {
```
RebeccaYo: Question: why did you change the text matching method from '~~' to grep?

doc/userDocs/usersGuide.md
```diff
@@ -141,7 +140,24 @@ The image below shows the lines that need to be edited in the configuration file
 2. Update *kubeconfigFile* to the full path to your kubeconfig file.  The paths will be identical if using a single cluster for both the app and driver.
 3. Update *kubeconfigContext* parameters with the name of the context for your cluster from the kubeconfig file, such as `kubernetes-admin@kubernetes`.
     - Set an empty string (`""`) to use the current-context defined in the kubeconfigFile.
-4. Set *useLoadBalancer* to `true` if your cluster supports LoadBalancer services. Otherwise, set it to `false` to use a NodePort for ingress.
+4. Choose a value for *appIngressMethod* that reflects the method you want to use
+  for communication between the workload drivers and the applications. You can choose 
+  from among:
+    - `clusterip`: Traffic from the workload drivers to the applications will use a
```
RebeccaYo: I notice later in the doc you adopt the ordering "loadbalancer", "nodeport" and "clusterip". I like having the default listed first and cluster IP last as it requires only one cluster and doesn't stress the entire network stack, so could you put clusterip last.

doc/userDocs/usersGuide.md
```diff
@@ -1356,4 +1403,79 @@ As a result, when using node labels it is recommended that you label all of your
...
+
+The command to start Minikube with 3 cpus and 8Gb is :
+* `minikube start --cpus=3 --memory=8g`
+
+Note that even with this cluster configuration, Minikube may have trouble running Weathervane 
+depending on how much memory is configured on the laptop and what else is 
+running.
+
+#### Weathervane Configuration
+
```
RebeccaYo: To run on Minikube, the user should still follow the Quickstart guide, correct? And the setup diverges when it comes to configuring Weathervane. For completeness, maybe add something like:
"To run Weathervane on Minikube, follow the Quickstart Setup steps Obtain Weathervane from GitHub and Create the Weathervane Docker Images. Then follow the instructions below to create the Weathervane Configuration file when running on Minikube."

doc/userDocs/usersGuide.md
hrosenbe: Yes.  Should have been "Note that ..."

doc/userDocs/usersGuide.md
hrosenbe: That ordering makes sense.  However the reason that I put clusterip first here is that it made it easier to explain that you might want to use loadbalancer or nodeport even on a single cluster.  

I will try putting clusterip last and rewording the others slightly.  Let me know what you think.

doc/userDocs/usersGuide.md
hrosenbe: Good idea.  I will add this.

hrosenbe: I actually added this in an overview section so that we don't have to repeat it for evey cluster type we describe here.

runHarness/AppInstance/AuctionAppInstance.pm
hrosenbe: I was having a strange issue with this test failing.  I thought that it might be due to some strange behavior of ~~, which is an experimental feature and is no longer really supported.  So I changed all of the ~~ matching in this file to use grep.  Then I realized that the test was using an otherwise unused parameter called _appInstanceSize_ rather than the correct _configurationSize_.  I must have used _appInstanceSize_ initially and then changed the name without deleting the parameter or changing it here.

PR #152, opened 2020-01-26
Force reload of data if data services don't start

hrosenbe:
* If the run harness is unable to start the data services for an app instance, it will delete the persistent volumes for that instance and retry.  This will force a reload of the data, which usually fixes errors that prevent instance from starting.
* Reduce the time spent waiting for services to start for non-backend services.  The current delay is necessary for warming large numbers of app servers, but is too long for the data services.
* When building the docker images, only build the executables if the image(s) being built will need them.  This speeds building just the run harness.

I tested this on a number of runs and it worked properly to catch and recover from a data corruption issue.


runHarness/DataManagers/AuctionKubernetesDataManager.pm
```diff
@@ -166,8 +160,18 @@ sub prepareDataServices {
...
+	if (!$allIsStarted) {
+		# Delete the PVCs for this namespace and try again
+		$console_logger->info(
+			"Couldn't start data services for appInstance $appInstanceNum of workload $workloadNum. Clearing data and retrying.\n" );
+		$appInstance->stopServices("data", $logPath);
+		my $cluster = $self->host;
+		my $namespace = $self->appInstance->namespace;
+		$cluster->kubernetesDeleteAllWithLabelAndResourceType("app=auction", "pvc", $namespace );
+		$allIsStarted = $appInstance->startServices("data", $logPath, 0);	
+	}
```
bhoflich: Should there be an option to turn this auto deleting off? Should it be default?
Maybe data services didn't start for some other reason and its a medium+ config size that takes a while to reload.

buildDockerImages.pl
```diff
@@ -233,23 +233,30 @@ sub cleanupAfterBuild {
 my $version = `cat version.txt`;
 chomp($version);
 
-# Build the executables
-print "Building the executables.\n";
-print $fileout "Building the executables.\n";
+# Build the executables in any of the images to be build
```
RebeccaYo: "if any of the images to be built"

runHarness/DataManagers/AuctionKubernetesDataManager.pm
hrosenbe: This makes sense.  I will add the option and have it off (no auto-delete) by default.  I will also update the documentation.

RebeccaYo: If it's off by default, @hrosenbe could you add a message to "Couldn't bring to running all data services for appInstance 12 of workload 1." that "If this problem recurs, you can enable auto-remediation by ..."?

hrosenbe: Good idea.  I will add this also.

buildDockerImages.pl
hrosenbe: fixed

PR #153, opened 2020-01-29
Add flexibility in namespace selection

hrosenbe:
Note that this pull request merges into the adjustProbes branch because it already contains the changes from that branch.  We will merge these changes into 2.0.3-dev with that branch.

The changes in this pull request are:
* Added parameters to allow control over the prefix and suffix of namespace names.
* Added a parameter to allow the specification of a list of namespace names.
* Added a parameter to control whether Weathervane tries to create namespaces.
* Updated the User's guide for the new parameters.
* Changed buildDockerImages.pl and build.gradle to use zip/unzip rather than tar.

### DIFF
--- a/doc/userDocs/usersGuide.md
+++ b/doc/userDocs/usersGuide.md
@@ -946,13 +946,14 @@ The default value is the empty string.
 The `namespaces` parameter specifies a list of namespace names to used by Weathervane 
 on this Kubernetes cluster. The value specified for this parameter must be a JSON list 
 of strings, each of which is a namespace name.  If a non-empty list is specified for 
-this parameter, then Weathervane will not generate namespace names for this cluster. 
-See [Configuring Namespaces](#namespaces) for more detail.
+this parameter, then Weathervane will ignore the values of `namespacePrefix` and `namespaceSuffix`, 
+and will not generate namespace names for this cluster. See [Configuring Namespaces](#namespaces) 
+for more detail.
   
 
-| Configuration Parameter: namespaceSuffix             |
+| Configuration Parameter: namespaces             |
 | ------------------------------------------------------ |
-| `"namespaceSuffix" : "",` |
+| `"namespaces" : "",` |
 
 The default value is an empty list.   
 
--- a/doc/userDocs/usersGuide.md
+++ b/doc/userDocs/usersGuide.md
@@ -1573,7 +1573,7 @@ and application namespaces.
     "name" : "k8sCluster", 
     "kubeconfigFile" : "/root/.kube/config",
     "kubeconfigContext" : "cluster-context-1",
-    "namespaces" : ["devNs1", "devNs2",],
+    "namespaces" : ["devns1", "devns2",],
   },
 ],
 ```
--- a/doc/userDocs/usersGuide.md
+++ b/doc/userDocs/usersGuide.md
@@ -1564,7 +1564,7 @@ generating the namespace names. Any values specified for `namespacePrefix` and `
 will be ignored.
 
 The following configuration file excerpt shows how to specify a list of namespace names. In 
-this example Weathervane will use *devNs1* and *devNs2* as the names for the workload driver 
+this example Weathervane will use *devns1* and *devns2* as the names for the workload driver 
 and application namespaces. 
 
 ```json
--- a/doc/userDocs/usersGuide.md
+++ b/doc/userDocs/usersGuide.md
@@ -271,110 +271,22 @@ namespaces are __auctionw1i*n*__, where __*n*__ is the application's instance nu
 name for the workload driver namespace is __auctionw1__.  Weathervane will try 
 to create these namespaces if they don't already exist.
 
-It is possible to customize the names for the namespaces, as well as whether Weathervane 
-should try to create the namespaces, for each Kubernetes cluster specified in your 
-configuration file. Additional details about the configuration parameters discussed 
-in this section are provided in 
-[Specifying Kubernetes Clusters](#kubernetesClusters) below.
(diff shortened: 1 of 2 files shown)


doc/userDocs/usersGuide.md
```diff
@@ -788,6 +905,73 @@ The `kubeconfigContext` parameter specifies the name of the context for your clu
...
+The `namespaces` parameter specifies a list of namespace names to used by Weathervane 
+on this Kubernetes cluster. The value specified for this parameter must be a JSON list 
+of strings, each of which is a namespace name.  If a non-empty list is specified for 
+this parameter, then Weathervane will not generate namespace names for this cluster. 
+See [Configuring Namespaces](#namespaces) for more detail.
+  
+
+| Configuration Parameter: namespaceSuffix             |
+| ------------------------------------------------------ |
+| `"namespaceSuffix" : "",` |
```
bhoflich: Should be "namespaces"x2 here.
Also, is there a note anywhere that "namespaces" takes precedence over Prefix and Suffix?

hrosenbe: I fixed the copy/paste error.

The text stating 

> If a non-empty list is specified for this parameter, then Weathervane will not generate namespace names for this cluster.

was supposed to indicate that the prefix and suffix would not be used, but I have made this more explicit.

doc/userDocs/usersGuide.md
```diff
@@ -259,6 +263,118 @@ but it will not stress the parts of the network stack that manage directing traf
...
+
+Weathervane runs each instance of the Auction application in a separate namespace. The 
+workload drivers also run in their own namespace.  The default names for the application 
+namespaces are __auctionw1i*n*__, where __*n*__ is the application's instance number.  The default 
+name for the workload driver namespace is __auctionw1__.  Weathervane will try 
+to create these namespaces if they don't already exist.
+
+It is possible to customize the names for the namespaces, as well as whether Weathervane 
+should try to create the namespaces, for each Kubernetes cluster specified in your 
+configuration file. Additional details about the configuration parameters discussed
```
RebeccaYo: This is a great place to discuss why you might want to customize the namespaces. 

I'm wondering if we can move this section to a less prominent location in the doc. It's currently in Detailed Setup Instructions. 

Specifying namespace names and disabling namespace creation seems to be a workaround for people who have namespace restrictions on their kubernetes cluster. Is this common? We could even create a new section 'advanced customizations' or 'special cases'.

Using the prefix and suffix, however, is geared to the typical user who wants to switch between configuration sizes without reloading the database, in other words, an advanced user, so I can see how it would fit in the Detailed Setup Instructions.

This section adds a lot of length to the doc. Could we possibly try to cut down on the length of this section as well.

doc/userDocs/usersGuide.md
```diff
@@ -259,6 +263,118 @@ but it will not stress the parts of the network stack that manage directing traf
...
+    "kubeconfigFile" : "/root/.kube/config",
+    "kubeconfigContext" : "cluster-context-1",
+    "namespaces" : ["devNs1", "devNs2",],
+  },
+],
+```
+When specifying namespace names, you must provide at least one namespace name for each 
+application instance you plan to run on the cluster, plus one namespace name for the 
+workload drivers.  This means that if you are running the drivers and applications 
+on the same cluster you will need to specify a minimum of two namespaces. In this case
```
RebeccaYo: "In this case the first namespace will be used for the workload drivers and others for the application instances"
Is the first namespace in the list always used for the driver and the second for application instances?, when one cluster is used for both? Does it alternate between driver and application, such as driver1, app1, driver2, app2 and so forth?

doc/userDocs/usersGuide.md
```diff
@@ -259,6 +263,118 @@ but it will not stress the parts of the network stack that manage directing traf
...
+
+The ability to specify a unique suffix is particularly useful when you are doing tests with different 
+configuration sizes on the same Kubernetes cluster.  Using a unique `namespaceSuffix` for 
+each configuration size will prevent the data services from being reloaded each time 
+you switch configuration size.
+
+##### Specifying Namespace Names
+
+There are situations in which namespace names must be used that do not fit the default or customized 
+name generation process.  Some examples are:
```
RebeccaYo: It would be good to note how common these kinds of restrictions are. Some users might be wondering "Do I have these restrictions or not, and do I have to research this, or should I just try running weathervane as-is"?

doc/userDocs/usersGuide.md
hrosenbe: I agree that this might be more detail than needed at this point in the doc.  I will move most of this content to the Advance Topics section.

doc/userDocs/usersGuide.md
hrosenbe: If the drivers and the apps are running on the same cluster, and use the same kubernetesCluster, then the first will always be used for the workload driver, and the others for the application instances.  Because we are not documenting the ability to have two workloads, there will never be more than one namespace used for workload driver.  If the driver and apps are running on different clusters, then there will be separate lists for each cluster.

Note that it is possible to specify two kubernetesCluster objects for the same cluster, with different names, and then give each a separate list of namespaces.  So you could have one cluster definition with a namespace for the driver, and another with namespaces for the applications, both pointing to the same kubeconfig and context.  I just didn't want to get into that much detail until we see that someone needs the functionality.  

There is no guarantee that the first namespace will be used for the drivers in all future versions though, so I think that it will be safest to take this line out.

hrosenbe:
Will open a new pull into 2.0.3-dev

PR #154, opened 2020-01-31
Add namespace flexibility and adjust probes

hrosenbe:
The changes in this pull request are:

* Added parameters to allow control over the prefix and suffix of namespace names.
* Added a parameter to allow the specification of a list of namespace names.
* Added a parameter to control whether Weathervane tries to create namespaces.
* Updated the User's guide for the new parameters.
* Changed buildDockerImages.pl and build.gradle to use zip/unzip rather than tar.
* Removed all liveness probes, as well as unneeded readiness probes.
* Increased the readiness probe period to 60 seconds.
* Randomized the initial delay on all probes.
* Removed the limit on RabbitMQ/Erlang threads


doc/userDocs/usersGuide.md
```diff
@@ -1418,6 +1519,93 @@ As a result, when using node labels it is recommended that you label all of your
...
+    "name" : "k8sCluster", 
+    "kubeconfigFile" : "/root/.kube/config",
+    "kubeconfigContext" : "cluster-context-1",
+    "namespaces" : ["devns1", "devns2",],
+  },
+],
+```
+When specifying namespace names, you must provide at least one namespace name for each 
+application instance you plan to run on the cluster, plus one namespace name for the 
+workload drivers.  This means that if you are running the drivers and applications
```
bhoflich: I would switch the order of these, since the driver namespace is allocated first? (per the comment in the other pr) 
workload drivers namespace then application instance namespaces.

hrosenbe: Agreed.  I have reworded.

PR #161, opened 2020-02-26
Add driver resource usage table to doc and fixedConfigSizes.xlsx

RebeccaYo:
Changelog:
- In fixedConfigSizes.xlsx, in sheet1, added a table for driver resource usage for multiple application instances. Previously, there was a table for application resource usage for multiple application instances, so now we have both driver and application usage.
- In fixedConfigSizes.xlsx, added a new sheet table3 which contains the table that is convertable to markdown format.
- In usersGuide.md, added table 3 which contains driver resource usage for multiple application instances.


RebeccaYo:
> It is fine the way it is, but the disk columns all being '0.00' aren't exactly necessary.
> Would something like a '- ' (dash/minus) work?

I agree it's a little clunky, but I wanted it to be as clear as possible that there's no disk requirement for the driver.

RebeccaYo:
Ben talked about this and we agree that 0.00 isn't the most accurate way to represent the disk requirements for the driver. There's the space that the images take up, which does not scale as the number of instances increases, and the PVC space, which is applicable only for the application requirements. There are different ways this could be addressed and Ben or I will file a doc bug for this. For now, we are ok with keeping it as is.

PR #163, opened 2020-03-16
Fix allocation of namespaces from namespace parameter

hrosenbe:
This pull request fixes the allocation of namespaces for the drivers and application instances so that namespaces are not allocated for the secondary drivers.  Currently a namespace from the namespaces parameter is consumed for each secondary driver, which is incorrect.


runHarness/WorkloadDrivers/AuctionKubernetesWorkloadDriver.pm
```diff
@@ -35,11 +35,13 @@ has 'namespace' => (
...
 override 'initialize' => sub {
-	my ( $self, $paramHashRef ) = @_;
+	my ( $self, $isMaster ) = @_;
 	super();
 
 	my $cluster = $self->host;
-	$self->namespace($cluster->kubernetesGetNamespace($self->workload->instanceNum, ""));
+	if ($isMaster) {
+		$self->namespace($cluster->kubernetesGetNamespace($self->workload->instanceNum, ""));
+	}
```
bhoflich: Will the secondaries work with an empty string for namespace?  Or is it set somewhere else?

hrosenbe: I added a line to set the namespace of a secondary driver to that of the master

PR #165, opened 2020-03-16
Allow workloadcontroller to contact stats service on localhost

hrosenbe:
Currently the workload controller uses its service name (wkldcontroller) to contact the stats service, which runs in the same container.  However, some Kubernetes networking providers do not allow traffic to hairpin to the source pod through its own service.  This prevents the workload controller from starting.

This change has the workload controller contact the stats service via the localhost hostname.  A separate workloadStatsHostname is used for the driver nodes so that they can still contact the stats service using the wkldcontroller service.


workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/core/Run.java
```diff
@@ -193,9 +195,6 @@ public void shutdown() {
 		 * Send exit messages to the other driver nodes
 		 */
 		for (String hostname : hosts) {
-			if (hostname == statsHost) {
-				continue;
-			}
```
bhoflich: Is it ok to completely remove this check?  Could it check for "localhost" on runStatsHost?

hrosenbe: I added back the check, but comparing against workloadStatsHost.  This check doesn't matter for Kubernetes, but for Docker it will matter if the workload controller also runs a driver node (which is the way it currently works).

PR #168, opened 2020-04-08
use context in AuctionKubernetesDataManager calls to kubectl

hrosenbe:
Changes in this pull request:
- The calls to kubectl in AuctionKubernetesDataManager were not using the context defined for the KubernetesCluster, and instead were just getting the current-context from the kubeconfig file.  This fixes that bug.

### DIFF
--- a/runHarness/DataManagers/AuctionKubernetesDataManager.pm
+++ b/runHarness/DataManagers/AuctionKubernetesDataManager.pm
@@ -298,10 +298,16 @@ sub loadData {
 	$logger->debug("Exec-ing perl /loadData.pl");
 	print $applog "Exec-ing perl /loadData.pl\n";
 	my $kubernetesConfigFile = $cluster->getParamValue('kubeconfigFile');
+	my $context = $cluster->getParamValue('kubeconfigContext');
+	my $contextString = "";
+	if ($context) {
+	  $contextString = "--context=$context";	
+	}
+
 	# Get the list of pods
 	my $cmd;
 	my $outString;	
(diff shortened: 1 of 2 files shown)


runHarness/DataManagers/AuctionKubernetesDataManager.pm
```diff
@@ -298,24 +298,30 @@ sub loadData {
...
 	$logger->debug("Command: $cmd");
 	$logger->debug("Output: $outString");
 	print $applog "Command: $cmd\n";
 	print $applog "Output: $outString\n";
 	my @lines = split /\s+/, $outString;
 	if ($#lines < 0) {
-		$console_logger->error("loadData: There are no pods with label auctiondatamanager in namespace $namespace");
-		exit(-1);
+		$console_logger->error("Data loading failed for Workload $workloadNum, appInstance $appInstanceNum: There are no pods with label auctiondatamanager in namespace $namespace");
+		return 0;
```
bhoflich, line 319: This is a big change to the error path.  Do tests show that it cleanly exits wv when this occurs?

hrosenbe: It is actually a bug fix, as the exit(-1) here resulted in Weathervane proceeding with the run despite the inability to load the data.  But your point is good that I need to make sure that this is the proper fix.  I will test that it exits properly when hitting this error.

hrosenbe: I tested this by killing the datamanager pod between the time it determined that the data wasn't loaded and the attempt to load the data.  The run exited cleanly with the appropriate error messages.

PR #169, opened 2020-04-22
Enable running multiple workloads

hrosenbe:
This pull request contains a number of fixes needed to enable running multiple Workloads.  
It also contains a fix to let weathervane work on a single node cluster, and uses the correct context in the data manager kubectl commands.

### DIFF
--- a/runHarness/WorkloadDrivers/AuctionWorkloadDriver.pm
+++ b/runHarness/WorkloadDrivers/AuctionWorkloadDriver.pm
@@ -1326,7 +1326,7 @@ sub startRun {
 									chop($successStr);
 								}
 								my $metricsStr = ", $successStr, throughput:$tptStr, avgRT:$rtStr";
-								$console_logger->info("   [appInstance: $appInstanceNum] Ended: $nameStr${metricsStr}.");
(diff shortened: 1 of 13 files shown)


runHarness/WorkloadDrivers/AuctionWorkloadDriver.pm
```diff
@@ -1326,7 +1326,7 @@ sub startRun {
 									chop($successStr);
 								}
 								my $metricsStr = ", $successStr, throughput:$tptStr, avgRT:$rtStr";
-								$console_logger->info("   [appInstance: $appInstanceNum] Ended: $nameStr${metricsStr}.");
+								$console_logger->info("   [workload $workloadNum, appInstance: $appInstanceNum] Ended: $nameStr${metricsStr}.");
```
bhoflich, line 1329: Should the workload info only print if num workloads > 1?  As this seems a somewhat unusual case.

hrosenbe: Yes.  Makes sense.  I will make this change before merging.

hrosenbe: Actually, I am going merge as-is and open a jira issue for this change.  It is non-trivial and I want to get these fixes in.

PR #171, opened 2020-04-27
Add small2 config

hrosenbe:
- Adds a small2 configuration.  This configuration is similar performance-wise to the existing small, but uses significantly less memory.  It also has CPU requests that are more in line with actual demands.
- Adds a parameter to runWeathervane.pl to enable mapping a user's .ssh directory.  This is needed when collecting esxtop.
- Adds a parameter that allows mapping in a different fixedConfigs.json file


runWeathervane.pl
```diff
@@ -32,7 +34,9 @@ package Weathervane;
 			'version=s' => \$version,
 			'tmpDir=s' => \$tmpDir,
 			'script=s' => \$backgroundScript,
+			'fixedConfigsFile=s' => \$fixedConfigsFile,
```
bhoflich, line 37: Does this open the possibility to achieve better scores with the same configurationsize by specifying a modified fixedConfigs file?

hrosenbe: Yes, but users can do that anyway by modifying the config file before building the runharness.  There is no way we keep users from "cheating" if they want to play with the configs.  Theoretically, the modified fixedConfigs file will end up in the output directory, but they could modify that also if they wanted.

PR #174, opened 2020-05-15
Add ability to make limits higher than requests

hrosenbe:
Changes in this pull request:

- Added the `limitsExpansionPct` parameter.  This allows the limits for the pods to be set to a certain percentage above the requests.  By default this only affects the appServer pod, but is `useKubernetesLimits` is set, it applies to all pods.  It is set to 0 by default.
- Updated the documentation to describe the use of requests and limits, along with the available parameters to change the behavior.


runHarness/Services/KubernetesService.pm
```diff
@@ -101,6 +101,41 @@ override 'isStopped' => sub {
 	return !$self->host->kubernetesDoPodsExist("type=$serviceType", $namespace );
 };
 
+sub expandK8sCpu {
+	my ($self, $cpuString, $expansionFactor) = @_;
+
```
bhoflich, line 106: Should there be some input validation on the expansionFactor string for this cpu and in the mem case? (numerical, >= 0)

hrosenbe: I added validation for the value of the limitsExpansionPct in AuctionAppInstance::checkConfig.  I tested the validation and it works.

PR #175, opened 2020-05-21
Explicitly delete deployments, etc rather than using all

hrosenbe:
Changes in this pull request:
- Rather than using `delete all` when cleaning up after a run, explicitly delete all constructs that might have been created by Weathervane.  Let Kubernetes deleted the pods.  Note that this change issues a delete for some kinds of objects (such as daemonsets) that are not currently used by Weathervane, but that may be in the future.

### DIFF
--- a/runHarness/ComputeResources/KubernetesCluster.pm
+++ b/runHarness/ComputeResources/KubernetesCluster.pm
@@ -314,49 +314,60 @@ sub kubernetesDeleteAllWithLabel {
 	if ($cmdFailed) {
 		$logger->error("kubernetesDeleteAllWithLabel delete deployments failed: $cmdFailed");
 	}
-	$logger->debug("Command: $cmd");
-	$logger->debug("Output: $outString");
-
+	if ($outString  && !($outString =~ /No\sresources\sfound/)) {
+		$logger->debug("Command: $cmd");
+		$logger->debug("Output: $outString");
+	}
+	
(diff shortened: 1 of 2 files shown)


runHarness/ComputeResources/KubernetesCluster.pm
```diff
@@ -309,13 +309,46 @@ sub kubernetesDeleteAllWithLabel {
...
+
+	$cmd = "kubectl delete service --selector=$selector --namespace=$namespace --kubeconfig=$kubeconfigFile $contextString";
+	($cmdFailed, $outString) = runCmd($cmd);
+	if ($cmdFailed) {
+		$logger->error("kubernetesDeleteAllWithLabel delete service failed: $cmdFailed");
+	}
+	$logger->debug("Command: $cmd");
+	$logger->debug("Output: $outString");
+
 	$cmd = "kubectl delete configmap --selector=$selector --namespace=$namespace --kubeconfig=$kubeconfigFile $contextString";
```
bhoflich, line 352: Doing it this way adds a lot of "No resources found" messages in the debug log due to not every selector/namespace having every type of resource to delete -- but that is ok.
Otherwise, seems to work without issues.

hrosenbe: I reduce the amount of logging by skipping the messages if the output is "No resource found".  There are still alot of runCmd log messages, but it is better.

PR #178, opened 2020-06-05
Add nodeport-internal as an appIngressMethod

hrosenbe:
Changes in this release:

- Added a new options for the `appIngressMethod` parameter called `nodeport-internal`.  This uses NodePort services for ingress from the workload drivers to the applications, but uses the Internal IP addresses of the nodes. This differs from the `nodeport` option which uses the External IP addresses of the nodes.  Note that in your drivers and applications are running in different clusters, then this option will only work if the nodes' internal addresses are routable from the driver cluster.
- Increase the timeouts on the readiness probes for most services to avoid occasional non-ready status when the nodes are actually OK.
- Change the Service type for RabbitMQ from NodePort to ClusterIP.  The use of NodePort was a bug that might cause an error on some clusters.
- Remove port 80 from the Nginx Service.  This port was not used, and it inclusion could use up load-balancer resources in some clusters.


doc/userDocs/usersGuide.md
```diff
@@ -152,11 +152,18 @@ The image below shows the lines that need to be edited in the configuration file
...
     - `nodeport`: Traffic from the workload drivers to the applications will
-          use the external IP addresses of the Kubernetes Nodes via a NodePort service. This 
+          use the *external* IP addresses of the Kubernetes Nodes via a NodePort service. This 
           may be desirable even when the workload drivers and applications are running 
           on the same cluster in order to load-test the complete networking 
           stack of your cluster. The cluster on which the applications run must 
           support exposing the node IP addresses externally using NodePort services.
+    - `nodeport-internal`: Traffic from the workload drivers to the applications will
+          use the *internal* IP addresses of the Kubernetes Nodes via a NodePort service. This 
+          may be desirable even when the workload drivers and applications are running
```
bhoflich, line 162: is the word 'even' needed here?

hrosenbe: I think so, since the main use of either nodeport or nodeport-internal will be when the drivers and applications are in different clusters, but you might want to use them even when they are in the same cluster.

PR #179, opened 2020-06-08
Don't use port 80 on nginx service if using https

hrosenbe:
Changes in this pull request:
- Release 2.0.5 removed port 80 from the nginx service, but there was an issue with the handling of the `ssl` parameter which is supposed to be able to turn off the use of https, in which case port 80 is needed.  The fix also introduced a spurious error message.  This pull request removes port 80 from the nginx service only if we are using https to communicate from the workload driver.


runHarness/AppInstance/AuctionKubernetesAppInstance.pm
```diff
@@ -155,7 +155,11 @@ override 'getEdgeAddrsRef' => sub {
 	  	 $logger->error("There are no IP addresses for the Kubernetes nodes");
 		 exit 1;
 	  }	
-      my $httpPort = $cluster->kubernetesGetNodePortForPortNumber("app=auction,type=webServer", 80, $self->namespace);
+	  # The http port will only be used if the ssl parameter is true
+      my $httpPort = 80;
+      if (!$self->getParamValue('ssl')) {
+      	$cluster->kubernetesGetNodePortForPortNumber("app=auction,type=webServer", 80, $self->namespace);
```
bhoflich, line 161: The return value can be ignored and just 80 used?

runHarness/AppInstance/AuctionKubernetesAppInstance.pm
```diff
@@ -155,7 +155,11 @@ override 'getEdgeAddrsRef' => sub {
 	  	 $logger->error("There are no IP addresses for the Kubernetes nodes");
 		 exit 1;
 	  }	
-      my $httpPort = $cluster->kubernetesGetNodePortForPortNumber("app=auction,type=webServer", 80, $self->namespace);
+	  # The http port will only be used if the ssl parameter is true
```
bhoflich, line 158: if the ssl parameter is false?

hrosenbe: Good point.  in that case we should remove port 443

runHarness/AppInstance/AuctionKubernetesAppInstance.pm
hrosenbe: No.  This is a bug.

runHarness/AppInstance/AuctionKubernetesAppInstance.pm
```diff
@@ -155,12 +155,16 @@ override 'getEdgeAddrsRef' => sub {
...
+	  # The http port will only be used if the ssl parameter is false
       my $httpPort = 80;
       if (!$self->getParamValue('ssl')) {
-      	$cluster->kubernetesGetNodePortForPortNumber("app=auction,type=webServer", 80, $self->namespace);
+      	$httpPort = $cluster->kubernetesGetNodePortForPortNumber("app=auction,type=webServer", 80, $self->namespace);
+      }
+	  # The https port will only be used if the ssl parameter is true
+	  my $httpsPort = 443;
+      if ($self->getParamValue('ssl')) {
+      	$httpPort = $cluster->kubernetesGetNodePortForPortNumber("app=auction,type=webServer", 443, $self->namespace);
```
bhoflich, line 166: Should it be $httpsPort =  (with the s)?

hrosenbe: yes

PR #181, opened 2020-06-19
doc updates  related to prereqs

bhoflich:
-In prerequisites, for StorageClass, add text to communicate/clarify that a Provisioner might have a number of required steps that need to be completed before running Weathervane. And, use vSphere Cloud Provider as an example with link.
-In prerequisites, for Perl, note that module Term::ReadKey is required.
-Adjust some quote characters for consistency.

### DIFF
--- a/doc/userDocs/usersGuide.md
+++ b/doc/userDocs/usersGuide.md
@@ -314,7 +314,7 @@ auction and item data, item images, and historical event data regarding bids
 and auction attendance.
 
 There must be at least one [StorageClass](#storageclass) defined on the application cluster to be used for this persistent storage.
-Examples for different storage providers are available at this link (https://kubernetes.io/docs/concepts/storage/storage-classes/).
+Examples for different storage providers are available at this link (https://kubernetes.io/docs/concepts/storage/storage-classes/).  
 A StorageClass Provisioner might have its own [prerequisites](#storageclass-prereq).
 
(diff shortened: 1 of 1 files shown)


doc/userDocs/usersGuide.md
```diff
@@ -94,9 +94,11 @@ Weathervane requires at least one existing Kubernetes cluster and a client syste
 - Kubernetes cluster:
     - You must have a kubeconfig file with [credentials](#credentials) for the cluster.
     - There must be at least one [StorageClass](#storageclass) defined on the cluster.
+        - A StorageClass Provisioner might have prerequisites and configuration [steps](#storageclass-prereq).
 - Client system:
     - Configure a workstation (MacOS or Linux) or a Linux VM with the required software:
         - Perl 5 (https://www.perl.org/get.html)
+        	- The module for Term::ReadKey should be installed.
```
hrosenbe, line 101: You should also include the command for this, e.g.:

The module for Term::ReadKey should be installed.  This can be achieved with the command `cpan Term::ReadKey`

bhoflich: The cpan command itself may not be sufficient because installing this module has other prerequisites, such as make and gcc.

hrosenbe: That is unfortunate. It would be good if we could remove the dependency on Term::ReadKey, but that should be a separate issue.  You might as well leave your text as-is for now.

PR #183, opened 2020-06-22
Add HTTP Proxy to buildDockerImages script

hrosenbe:
This pull request adds a proxy parameter that allows an http(s) proxy to be specified for the buildDockerImages script.


dockerImages/zookeeper/Dockerfile
```diff
@@ -1,17 +1,19 @@
 # Copyright 2017-2019 VMware, Inc.
 # SPDX-License-Identifier: BSD-2-Clause
-FROM namespace/weathervane-centos7:version
+FROM hrosenbe/weathervane-centos7:2.0.7-dev
```
bhoflich, line 3: hrosenbe

hrosenbe: Good catch.  Thanks.

PR #184, opened 2020-07-07
Fix Interval Run Strategy

hrosenbe:
Changes in this pull request:
- The Interval Run Strategy, which allows varying loads, had not been updated properly for changes from Weathervane 1.0 to 2.0 and did not work properly.  The changes in this pull request fix the interval load path so that it works as in Weathervane 1.0.  Note that this pull request does not update the documentation, so this is still an undocumented and unsupported feature.  The documentation will be updated for the next release once we do more testing with this feature.
- Fixed a bug in the Auction application that resulted in an exception when trying to retrieve a full-sized image for non-current items that do not have full images loaded.

### DIFF
--- a/auctionApp/src/main/java/com/vmware/weathervane/auction/data/imageStore/ImageStoreFacadeCassandraImpl.java
+++ b/auctionApp/src/main/java/com/vmware/weathervane/auction/data/imageStore/ImageStoreFacadeCassandraImpl.java
@@ -255,10 +255,12 @@ public class ImageStoreFacadeCassandraImpl extends ImageStoreFacadeBaseImpl {
 	public byte[] retrieveImage(UUID imageHandle, ImageSize size) throws NoSuchImageException,
 			IOException {
 		logger.info("retrieveImage imageHandle = " + imageHandle + ", imageSize = " + size);
+		List<ImageThumbnail> thumbs = imageThumbnailRepository.findByKeyImageId(imageHandle);
+		List<ImagePreview> previews = imagePreviewRepository.findByKeyImageId(imageHandle);
+		List<ImageFull> fulls = imageFullRepository.findByKeyImageId(imageHandle);
 		switch (size) {
 		case THUMBNAIL:
-			List<ImageThumbnail> thumbs = imageThumbnailRepository.findByKeyImageId(imageHandle);
-			if (thumbs == null) {
+			if ((thumbs == null) || (thumbs.size() <= 0)) {
 				logger.warn("retrieveImage thumbs = null, imageHandle = " + imageHandle
 						+ ", imageSize = " + size);
 				throw new NoSuchImageException();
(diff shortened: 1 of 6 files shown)


auctionApp/src/main/java/com/vmware/weathervane/auction/data/imageStore/ImageStoreFacadeCassandraImpl.java
```diff
@@ -255,33 +255,44 @@ public void addImages(List<ImageInfo> imageInfos, List<BufferedImage> fullImages
 	public byte[] retrieveImage(UUID imageHandle, ImageSize size) throws NoSuchImageException,
 			IOException {
 		logger.info("retrieveImage imageHandle = " + imageHandle + ", imageSize = " + size);
+		List<ImageThumbnail> thumbs = imageThumbnailRepository.findByKeyImageId(imageHandle);
+		List<ImagePreview> previews = imagePreviewRepository.findByKeyImageId(imageHandle);
+		List<ImageFull> fulls = imageFullRepository.findByKeyImageId(imageHandle);
 		switch (size) {
```
bhoflich, line 261: Will retrieving all 3 images vs just 1 depending on the case have a performance impact?

runHarness/WorkloadDrivers/AuctionWorkloadDriver.pm
```diff
@@ -1309,6 +1314,12 @@ sub startRun {
 							if (!($endIntervalName =~ /InitialRamp\-(\d+)/)) {
 								my $tptStr = sprintf("%.2f", $statsSummary->{"throughput"});
 								my $rtStr = sprintf("%.2f", $statsSummary->{"avgRT"});
+								my $startUsersStr = $statsSummary->{"startActiveUsers"};
+								my $endUsersStr = $statsSummary->{"endActiveUsers"};
+								my $numRtOps = $statsSummary->{"totalNumRTOps"};
+								my $numFailedRt = $statsSummary->{"totalNumFailedRT"};
+								my $pctFailRTStr = sprintf("%.2f", 100 * ((1.0 * $numFailedRt) / $numRtOps));
+
```
bhoflich, line 1322: Should there be a zero check on numRtOps?

auctionApp/src/main/java/com/vmware/weathervane/auction/data/imageStore/ImageStoreFacadeCassandraImpl.java
hrosenbe: Yes.  Good catch.  I will change this.

hrosenbe: fixed

runHarness/WorkloadDrivers/AuctionWorkloadDriver.pm
hrosenbe: Yes.  Fixed

PR #185, opened 2020-07-08
For xsmall, increase nginx cache disk and decrease percentage used

hrosenbe:
Changes in this pull request:
- To ensure that the nginx cache does not fill its disk, decrease the percentage of the disk used for the xsmall config to 85% of the disk size.
- To ensure that the decreased percentage doesn't affect performance, increase the size of the nginx disk by 5% to 2150Mi.
- update the documentation to reflect the change


doc/userDocs/usersGuide.md
```diff
@@ -1039,34 +1039,34 @@ instances you will be able to deploy on your clusters.
...
+| 15                                    |  11\.87  |  83\.50           |  480\.00        |  38\.85  |  187\.65          |  886\.00        |  72\.00  |  477\.54          |  1,200\.00      |
+| 16                                    |  12\.66  |  89\.06           |  512\.00        |  41\.44  |  200\.16          |  944\.20        |  76\.80  |  509\.38          |  1,280\.00      |
+| 17                                    |  13\.45  |  94\.63           |  544\.00        |  44\.03  |  212\.67          |  1,003\.40      |  81\.60  |  541\.21          |  1,360\.00      |
+| 18                                    |  14\.24  |  100\.20          |  576\.00        |  46\.62  |  225\.18          |  1,062\.60      |  86\.40  |  573\.05          |  1,440\.00      |
+| 19                                    |  15\.03  |  105\.76          |  608\.00        |  49\.21  |  237\.69          |  1,121\.80      |  91\.20  |  604\.88          |  1,520\.00      |
+| 20                                    |  15\.82  |  111\.33          |  640\.00        |  51\.80  |  250\.20          |  1,181\.00      |  96\.00  |  636\.72          |  1,600\.00      |
```
bhoflich, line 1069: That is a lot of 0.2 multiples caused by rounding 0.15. (due to 2150).

runHarness/fixedConfigs.json
```diff
@@ -138,7 +138,7 @@
   "webServerCpus" : "275m",
   "webServerMem"  : "2Gi",
   "webServerImpl" : "nginx",
-  "nginxCacheVolumeSize" : "2Gi",
+  "nginxCacheVolumeSize" : "2150Mi",
```
bhoflich, line 142: Its such a small change, but should a xsmall2 instead of xsmall be considered?

doc/userDocs/usersGuide.md
hrosenbe: Yes.  The additional size is relatively small so I considered ignoring it, but I decided to err on the side of ensuring that we don't under-report the requirements.

runHarness/fixedConfigs.json
hrosenbe: Since it is such a small change, and since xsmall is so new that it is unlikely it has been used, I think that we should just keep it called xsmall.  It seems too confusing to add small2 for a change that won't affect performance.

PR #187, opened 2020-07-29
Ensure Datamanager pod is stopped before starting other pods

hrosenbe:
The AuctionDataManager pod takes a large amount of resources (CPU and memory) relative to other pods.  Currently the prepareData process can return before this pod has terminated. This can keep other pods from starting, leading to misleading failedScheduling errors.

The fix is to wait until the pod has terminated before returning.

### DIFF
--- a/runHarness/DataManagers/AuctionKubernetesDataManager.pm
+++ b/runHarness/DataManagers/AuctionKubernetesDataManager.pm
@@ -141,6 +141,11 @@ sub stopDataManagerContainer {
 	$cluster->kubernetesDelete("configMap", "auctiondatamanager-config", $self->appInstance->namespace);
 	$cluster->kubernetesDelete("deployment", "auctiondatamanager", $self->appInstance->namespace);
 
+	# Don't return until the data manager pod has terminated
+    my $podExists = 1;
+    do {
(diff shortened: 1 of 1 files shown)


runHarness/DataManagers/AuctionKubernetesDataManager.pm
```diff
@@ -141,6 +141,11 @@ sub stopDataManagerContainer {
 	$cluster->kubernetesDelete("configMap", "auctiondatamanager-config", $self->appInstance->namespace);
 	$cluster->kubernetesDelete("deployment", "auctiondatamanager", $self->appInstance->namespace);
 
+	# Don't return until the data manager pod has terminated
+    my $podExists = 1;
+    do {
+	  $podExists = $self->host->kubernetesDoPodsExist("impl=auctiondatamanager", $self->appInstance->namespace );    	
+    } while ($podExists);
 }
```
bhoflich, line 149: Should this have a small sleep in the loop and a max retries?

hrosenbe: Yes and yes.  I will add these

hrosenbe: I added a sleep and max retries.  Even if the pod hasn't terminated after the retries I let the run proceed.  usually this won't be an issue.

PR #196, opened 2020-10-15
add AppServerWarmer for docker only

bhoflich:
to be consistent with the k8s version which already has the warmer

### DIFF
--- a/runHarness/Services/TomcatDockerService.pm
+++ b/runHarness/Services/TomcatDockerService.pm
@@ -18,6 +18,12 @@ with Storage( 'format' => 'JSON', 'io' => 'File' );
 
 extends 'Service';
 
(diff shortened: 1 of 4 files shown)


runHarness/Services/TomcatDockerService.pm
```diff
@@ -183,6 +230,17 @@ sub isUp {
...
+			return 1;
+		}
+	}
+	else {
+		return 0;
+	}
+
+	my $warmerTarget = $self->warmerTarget;
+	$response = `curl -s http://$warmerTarget/warmer/ready`;
+	if ( $response =~ /ready/ ) {
```
hrosenbe, line 243: For debugging purposes it would be useful to log the url and response here.

hrosenbe (approved):

PR #198, opened 2020-11-16
Enable overriding of dockerNamespace on command-line

hrosenbe:
- It is now possible to override the dockerNamespace from the weathervane config file by specifying it on the command-line to runWeathervane.pl.  Because this is overriding a config file parameter, the parameter must be specified are `--` on the command-line as in: 

`./runWeathervane.pl --configFile=weathervane.config.xsmall -- --dockerNamespace=myNamespace`

### DIFF
--- a/runWeathervane.pl
+++ b/runWeathervane.pl
@@ -103,14 +103,13 @@ sub usage {
 }
 
 sub parseConfigFile {
-	my ($configFileName) = @_;
+	my ($configFileName, $dockerNamespace) = @_;
 	
 	# Read in the config file
 	open( my $configFile, "<$configFileName" ) or die "Couldn't open configuration file $configFileName: $!\n";
 
 	my @k8sConfigFiles;
 	my %clusterNameToKubeconfig;
-	my $dockerNamespace;
 	my $topLevelAppInstanceCluster = "";
 	my %topLevelStorageClassNames;
 	# Parsing the config file manually to avoid requiring the JSON package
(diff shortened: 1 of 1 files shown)


runWeathervane.pl
```diff
@@ -449,9 +447,16 @@ sub forceLicenseAccept {
     $sshMountString = "-v $homeDir/.ssh:/root/.ssh";
 }
 
-my $retRef = parseConfigFile($configFile);
+my $dockerNamespace;
+# Determine whether the dockerNamespace is set in the command-line
+# options that are getting passed to the runHarness
+if ($wvCommandLineArgs =~ /\-\-dockerNamespace(\=|\s)([^\s]*)(\s|$)/) {
```
bhoflich, line 453: Would it be valid to have multiple spaces following dockerNamespace when not using '='?
--dockerNamespace  (multiple spaces here)   myName
if so, 1st capturing group could maybe adjust (\=|\s) to (\=|\s+)    ?

bhoflich (approved):

RebeccaYo (approved):
I tested this change on my setup and it looked fine. It's great to be able to specify this from the command line, thank you.

runWeathervane.pl
hrosenbe: Actually, the way that perl puts arguments into ARGV ignores extra spaces, and the way that runWeathervane.pl puts them back together into a single string, ensures that there will never be multiple consecutive spaces in $wvCommandLineArgs.

PR #217, opened 2021-06-04
Bhoflich/doc updates 2.1

bhoflich:
-doc updates for pod affinity and anti-affinity
-doc updates to 2.1
-remove small2-applimit1
-update fixedConfigsSizes.xlsx and tables


doc/userDocs/usersGuide.md
```diff
@@ -1551,6 +1550,21 @@ As a result, when using node labels it is recommended that you label all of your
 worker nodes with either `wvrole=driver` or `wvrole=sut`.  This will give you the 
 best control over pod placement.
 
+### Inter-pod affinity and anti-affinity<a name="pod-affinity"></a>
+
+With the Weathervane 2.1 release, configuration parameters were provided to manage the addition of pod affinity and anti-affinity rules with weights to the Weathervane pod manifests. 
+This provides more control to which nodes the pods are eligible to be scheduled on.
```
hrosenbe, line 1556: Maybe "more control for which nodes"

doc/userDocs/usersGuide.md
```diff
@@ -1551,6 +1550,21 @@ As a result, when using node labels it is recommended that you label all of your
...
+
+| Parameter:                      | Description: |
+| ------------------------------- | ------------ |
+| `serviceTypeAntiAffinity`       | If true, a podAntiAffinity rule will be used to create anti-affinity among pods of the same type across all application instances. (Default: true) |
+| `serviceTypeAffinity`           | If true, podAffinity and podAntiAffinity rules will be used to create affinity among pods of the same type and anti-affinity among pods of different types.  These will apply across all applications instances. (Default: false) |
+| `podInstanceAffinity`           | If true, a podAffinity rule will be used to create affinity among pods of the same application instance. (Default: false) |
+| `podInstanceAffinityWeight`     | The weight to apply to the podAffinity rule if podInstanceAffinity is true. (Default: 50) |
+| `serviceTypeAntiAffinityWeight` | The weight to apply to the podAntiAffinity rule if serviceTypeAntiAffinity is true. (Default: 100) |
+
+Weathervane defaults to use serviceTypeAntiAffinity, as this leads to a more even load across all nodes and more predictable results.
```
hrosenbe, line 1566: There should be note here about which combinations are incompatible.  Something like:

Note that there are natural restrictions on the combinations of these variables which are allow to be true on a given run.  Only one of  serviceTypeAntiAffinity, serviceTypeAffinity, and podInstanceAffinity can be true on a given run.

hrosenbe (approved):

doc/userDocs/usersGuide.md
```diff
@@ -1551,6 +1550,23 @@ As a result, when using node labels it is recommended that you label all of your
 worker nodes with either `wvrole=driver` or `wvrole=sut`.  This will give you the 
 best control over pod placement.
 
+### Inter-pod affinity and anti-affinity<a name="pod-affinity"></a>
+
+With the Weathervane 2.1 release, configuration parameters were provided to manage the addition of pod affinity and anti-affinity rules with weights to the Weathervane pod manifests. 
+This provides more control for which nodes the pods are eligible to be scheduled on.
```
RebeccaYo, line 1556: "which nodes the pods are eligible to be scheduled on": Is the affinity setting a "hint/best effort" or a requirement for pod scheduling? If it's the latter, are there implications for the kubernetes cluster, as in number of nodes, size of nodes, etc and is that the reason for the change to the Application Resource Requirements table? Does the kubernetes resource requirement change if a non-default affinity setting is selected?

runHarness/Parameters.pm
```diff
@@ -1143,8 +1143,8 @@ $parameters{"serviceTypeAffinity"} = {
 	"type"      => "!",
 	"default"   => JSON::false,
 	"parent"    => "runManager",
-	"usageText" => "If true, podAffinity and podAntiAffinity rules will be used to create affinity among pods of the same type across all application instances,\n" .
-	" and anti-affinity among pods of different types across all instances",
+	"usageText" => "If true, podAffinity and podAntiAffinity rules will be used to create affinity among pods of the same types,\n" .
+	" and anti-affinity among pods of different types. These will apply across all applications instances",
```
RebeccaYo, line 1147: Should applications be application?

doc/userDocs/usersGuide.md
bhoflich: The kubernetes docs use the terms "should" and "should not", which I take to mean best-effort.
Either way there are implications because packing the boxes (nodes) differently may result in the same contents (pods) no longer able to fit or able to fit and leaving different sized spaces that other stuff can now fit into. The summed value resource requirement should not change with different affinity, but if 'fixed-size' pods no longer fit due to 'fragmentation' of the node available resources, more resources may be required.  There may also be performance implications due to locality.  
I changed the application resources table because the values were not correct anymore, and didn't match the spreadsheet values (before or after I made minor updates).

doc/userDocs/usersGuide.md
RebeccaYo: Recommend moving this line to the first paragraph. I think that's one of the first things that should be said about affinity - why it was added

doc/userDocs/usersGuide.md
```diff
@@ -1551,6 +1550,23 @@ As a result, when using node labels it is recommended that you label all of your
...
+| Parameter:                      | Description: |
+| ------------------------------- | ------------ |
+| `serviceTypeAntiAffinity`       | If true, a podAntiAffinity rule will be used to create anti-affinity among pods of the same type across all application instances. (Default: true) |
+| `serviceTypeAffinity`           | If true, podAffinity and podAntiAffinity rules will be used to create affinity among pods of the same type and anti-affinity among pods of different types.  These will apply across all application instances. (Default: false) |
+| `podInstanceAffinity`           | If true, a podAffinity rule will be used to create affinity among pods of the same application instance. (Default: false) |
+| `podInstanceAffinityWeight`     | The weight to apply to the podAffinity rule if podInstanceAffinity is true. (Default: 50) |
+| `serviceTypeAntiAffinityWeight` | The weight to apply to the podAntiAffinity rule if serviceTypeAntiAffinity is true. (Default: 100) |
+
+Note there are natural restrictions on the combination of these parameters on a given run. The serviceTypeAffinity rule can not be true when serviceTypeAntiAffinity or podInstanceAffinity are true.
+
```
RebeccaYo, line 1567: Why might a user be interested in setting serviceTypeAffinity, podInstanceAffinity? Why might they need to change the weight? Why were they added to the benchmark? Or likewise, should users not usually have to modify these policies?

doc/userDocs/usersGuide.md
```diff
@@ -1551,6 +1550,23 @@ As a result, when using node labels it is recommended that you label all of your
...
+This provides more control for which nodes the pods are eligible to be scheduled on.
+
+| Parameter:                      | Description: |
+| ------------------------------- | ------------ |
+| `serviceTypeAntiAffinity`       | If true, a podAntiAffinity rule will be used to create anti-affinity among pods of the same type across all application instances. (Default: true) |
+| `serviceTypeAffinity`           | If true, podAffinity and podAntiAffinity rules will be used to create affinity among pods of the same type and anti-affinity among pods of different types.  These will apply across all application instances. (Default: false) |
+| `podInstanceAffinity`           | If true, a podAffinity rule will be used to create affinity among pods of the same application instance. (Default: false) |
+| `podInstanceAffinityWeight`     | The weight to apply to the podAffinity rule if podInstanceAffinity is true. (Default: 50) |
+| `serviceTypeAntiAffinityWeight` | The weight to apply to the podAntiAffinity rule if serviceTypeAntiAffinity is true. (Default: 100) |
```
RebeccaYo, line 1564: Since serviceTypeAntiAffinity comes first in the list (and it's the default), serviceTypeAntiAffinityWeight should come before podInstanceAffinityWeight

doc/userDocs/usersGuide.md
RebeccaYo: "This provides more control for which nodes the pods are eligible to be scheduled on" -> recommended "This changes how the Weathervane pods are scheduled across Kubernetes nodes"

In regards to the best effort vs requirement question, the Kubernetes documentation [here](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity) states "you can indicate that the rule is "soft"/"preference" rather than a hard requirement, so if the scheduler can't satisfy it, the pod will still be scheduled;" so it sounds like this is something you can set. Are the affinity labels requiredDuringSchedulingIgnoredDuringExecution and preferredDuringSchedulingIgnoredDuringExecution relevant here?

Great explanation Ben. Based on this explanation, here are a few ideas of things that sound like they should be included:
"Changing the affinity setting will result in noncomparable performance results."
"Changing the affinity setting could result in different resource requirements at the Kubernetes node level, such as different CPU or memory resources requirements per node or requiring a different number of nodes to run the same number of Weathervane Application Instances."

doc/userDocs/usersGuide.md
bhoflich: They are mostly for experimentation.  We used them to understand the performance impacts.
We had a least one external request for serviceTypeAffinity functionality.
Normal users won't typically modify these.

doc/userDocs/usersGuide.md
bhoflich: Updated. Does this look better?

doc/userDocs/usersGuide.md
```diff
@@ -83,9 +83,18 @@ fixed amount of throughput in terms of both operations-per-second and
 HTTP-requests/sec.  As a result, there is a direct correlation between WvUsers
 and throughput.
 
+
+### Weathervane 2.1 vs 2.0
+
+The changes in Weathervane 2.1 may be significant enough to slightly alter 
+the performance results vs Weathervane 2.0 baselines.  It is advised not to directly
```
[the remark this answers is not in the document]
bhoflich: Updated.

doc/userDocs/usersGuide.md
RebeccaYo: The changes you made look good to me, thank you. It would be nice to have a sentence explaining to users why they would or wouldn't use these parameters, such as "These parameters will not need to be modified in typical use." But it's not a must have for me.

bhoflich: Done

RebeccaYo (approved):
The question about whether the affinity rules are best effort or requirements - still an open question for me, but it's probably not necessary to cover these implementation details in the doc, especially considering the parameters are really just for experimentation purposes

PR #219, opened 2021-07-09
Bhoflich/interval run strategy fix

bhoflich:
This fixes an error for uninitialized value $aiPassed... when using the interval runStrategy.
It also adds the successStr into the print message for the same runStrategy, to make them more meaningful.

I am pull requesting this into master because it is a bug fix that does not impact performance.


RebeccaYo (approved):
I can see how
$self->passRT->{$appInstanceNum} = 0;
$self->passFailure->{$appInstanceNum} = 0;
were initialized in the case that max passing interval could not be found, but were never initialized in the case that maxPassInterval never existed due to use of Interval strategy which doesn't find max passing users. Then, those values are called in isPassed.
I think it's fine to merge this into master too.

PR #228, opened 2022-02-01
collect k8s events log

lalvarezsanc:
Collecting kubernetes event log and storing in weathervane/output/<run_num>/0/configuration/clusters/k8sCluster.


runHarness/ComputeResources/KubernetesCluster.pm
```diff
@@ -148,6 +148,19 @@ override 'getConfigFiles' => sub {
         print FILEOUT "$line\n";
     }
     close FILEOUT;
+				
+    $cmd = "kubectl get events --sort-by=.metadata.creationTimestamp";
```
bhoflich, line 152: Shouldn't the kubeconfig and contextString also be provided in the command?
--kubeconfig=$kubeconfigFile $contextString
Also, should probably get the events for all the namespaces to enable more info:
 --all-namespaces

lalvarezsanc: Nice catch

PR #230, opened 2022-02-03
cleanup after building images

lalvarezsanc:
Calling docker rmi -f $(docker images -q) and docker container prune -f within cleanupAfterBuild subroutine.

### DIFF
--- a/buildDockerImages.pl
+++ b/buildDockerImages.pl
@@ -199,11 +199,22 @@ sub setupForBuild {
 	runAndLog($fileout, "cp -r ./workloadConfiguration ./dockerImages/runharness/workloadConfiguration");
 }
 
+sub removeImagesAndContainers {
+	my ($fileout, @imageIDs) = @_;
+	
+	foreach my $imageID (@imageIDs){
+		runAndLog($fileout, "docker rm \$(docker ps -a | grep \"${imageID}\" | awk '{print \$1}')"); # Removing containers
+		runAndLog($fileout, "docker images -a | grep \"${imageID}\" | awk '{print \$3}' | xargs docker rmi"); # removing image
+	}
+	#Catching any left-over images/containers
+	runAndLog($fileout, "docker images -a | grep \"weathervane*\\|openjdk*\\|centos*\" | awk '{print \$3}' | xargs docker rmi"); # removing images
+	runAndLog($fileout, "docker rm \$(docker ps -a | grep \"weathervane*\" | awk '{print \$1}')"); # Removing containers
+}
+
 sub cleanupAfterBuild {
 	my ($fileout, @imageIDs) = @_;
 	#cleaning extraneous files from previous runs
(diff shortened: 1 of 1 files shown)


buildDockerImages.pl
```diff
@@ -195,6 +195,10 @@ sub setupForBuild {
 
 sub cleanupAfterBuild {
 	my ($fileout) = @_;
+	#cleaning extraneous files from previous runs
+	runAndLog($fileout, "docker rmi -f \$(docker images -q)");
+	runAndLog($fileout, "docker container prune -f");
```
bhoflich, line 200: The first line removes all docker images including non weathervane related ones?
The second line removes all containers including non weathervane related ones?
I don't think every user wants to do either of these by default and maybe not even optionally as part of the script -- they can always been done manually if needed.
Is there a straight forward way to prune by some filter just the weathervane related containers typically left behind?

lalvarezsanc: That's a good point, the same thought crossed my mind. I'll see if there's a way to filter which gets deleted and which doesn't.

RebeccaYo:
@lalvarezsanc @bhoflich 
Hi Luis, I apologize for not getting to this yesterday, I had an urgent issue come up.

I tested "docker images -a | grep \"weathervane*\" | awk '{print \$3}' | xargs docker rmi" on the CICD driver VM. Afterward, I found some remaining images that were not named 'weathervane-' but were still used during the build.

Before running the cleanup command:
image

After running cleanup command: 
image

You might consider if you want to specify these for deletion.

The CICD pipeline where this was tested specifies to download openjdk and harbor-repo.vmware.com/dockerhub-proxy-cache/library/openjdk images from Corporate Harbor because if not specified, it was defaulting to download from DockerHub which was failing. If the default DockerHub download was succeeding, I'm not sure whether the result of cleanup would be the same or not, that is, you would want to test what the openjdk repository and tag would read.

Regarding the container cleanup command, did you confirm that the container created on the driver during build can be filtered using "weathervane*"?
Here's the containers that were present on my Driver prior to cleanup, notice the names are not specified:
image

RebeccaYo:
> However, I do not seem to get any leftover containers after the running buildDockerImages script

I believe the container is left over when the script is aborted midway. I wouldn't expect it to be left on normal execution.

buildDockerImages.pl
```diff
@@ -193,8 +194,18 @@ sub setupForBuild {
 	runAndLog($fileout, "cp -r ./workloadConfiguration ./dockerImages/runharness/workloadConfiguration");
 }
 
+sub removeImagesAndContainers {
```
RebeccaYo, line 197: Only one more comment; since the container removal is no longer present, this subroutine should be renamed.

RebeccaYo (changes requested):

RebeccaYo (approved):

PR #237, opened 2022-03-17
small3 config size

lalvarezsanc:
Changes:
* Added small3 configuration size
* Warning if a deprecated configuration size is used


lalvarezsanc:
Note: I Haven't been able to fully test the new configuration size due to system complications.

lalvarezsanc:
Deprecated Warning
This is what a deprecated warning looks like—there is only one warning now instead of two.

bhoflich (approved):

RebeccaYo (approved):

PR #238, opened 2022-03-23
update email addresses

bhoflich:
remove all old email addresses and replace with weathervane-dev where appropriate


lalvarezsanc (approved):
Looks good

PR #239, opened 2022-04-15
increasing micro app warmer cpus

lalvarezsanc:
Tested both 150m and 100m two times. Here's how long it took to start the backend services for each one:

**150**
* 17 minutes
* 14 minutes

**100**
* 15 minutes
* 14 minutes

Because both 100m and 150m produced similar results, I changed the appWarmerCpus parameter to 100m.


bhoflich:
How long did it take with the unchanged value ("50m") ?

bhoflich (approved):

PR #252, opened 2023-04-27
change readerwriter lock to simple sync object lock

bhoflich:
It does not appear the current implementation in PerWorkloadStatsColletor using a reader writer lock is correct.

This results in log error messages like this:
0/0/logs/workloadDriver/driverCluster/wklddriver-3.log:xxxxx [http-nio-80-exec-6] WARN  c.v.w.w.c.s.s.PerWorkloadStatsCollector - statsIntervalComplete: caught null operationStats.

Updated to use simple locking with a synchronized block on an object.

Also updated the version.txt to match the branch 2.1.2-dev


workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/statistics/statsCollector/PerWorkloadStatsCollector.java
```diff
@@ -93,22 +92,12 @@ public PerWorkloadStatsCollector(List<StatsIntervalSpec> statsIntervalSpecs, Loa
...
+			synchronized (curStatsListLock) {
 				curStatsList.add(operationStats);
-			} else {
-				logger.warn("submitOperationStats for workload {}, received null operationStats", workloadName);
 			}
-		} finally {
-			curStatsRWLock.readLock().unlock();
-			logger.debug("submitOperationStats unlocked readLock: " + operationStats);
+		} else {
+			logger.warn("submitOperationStats for workload {}, received null operationStats", workloadName);
```
RebeccaYo, line 100: Should this be an error vs a warning? operationStats shouldn't be null

RebeccaYo: And likewise for the other null operationStats warning below (which I can't leave a comment on)

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/statistics/statsCollector/PerWorkloadStatsCollector.java
```diff
@@ -93,22 +92,12 @@ public PerWorkloadStatsCollector(List<StatsIntervalSpec> statsIntervalSpecs, Loa
 	public void submitOperationStats(OperationStats operationStats) {
 		logger.debug("submitOperationStats: " + operationStats);
```
RebeccaYo, line 94: Add comment back in

RebeccaYo (approved):
See comments below; either way this is approved. Ben and I talked today to discuss the changes.

workloadDriver/src/main/java/com/vmware/weathervane/workloadDriver/common/statistics/statsCollector/PerWorkloadStatsCollector.java
bhoflich: I think it's the same either way for now, since the logger setting is warn. leaving it warn for now -- should probably just remove in future since it (in theory) can't happen).
I don't want to touch the one below since I didn't study the code and don't want to break anything.

Issue #251, opened 2023-04-25
The kubernetesDeleteAllForCluster subroutine is trying to delete all config maps, failing on OpenShift clusters


krisiasty:
### Describe the bug

The `kubernetesDeleteAllForCluster `subroutine is trying to delete all config maps (without any selector), which also includes system-generated configmaps like `kube-root-ca.crt` and `openshift-service-ca.crt` on OpenShift clusters. These configmaps are managed by cluster operators and are recreated shortly after each modification or deletion, so when weathervane is trying to check whether all resources were deleted it still finds those configmaps and eventually fails after several iterations.

### Reproduction steps

1. Deploy and run weathervane 2.1.1 on Red Hat OpenShift cluster (tested on OCP 4.12)
2.
3.
...

### Expected behavior

Only resources with properly set label (by default `app=auction`) should be deleted and checked.

I found in the [KubernetesCluster.pm](https://github.com/vmware/weathervane/blob/master/runHarness/ComputeResources/KubernetesCluster.pm) source code there are several calls to `kubectl` with the `--selector=app=auction` argument, but for some reason this is missing in the `kubernetesDeleteAllForCluster` subroutine.

bhoflich:
This is fixed in master with the 2.1.2 release.

PR #253, opened 2023-04-27
use selector in DeleteAllForCluster

bhoflich:
to fix issue #251 
The kubernetesDeleteAllForCluster subroutine is trying to delete all config maps, failing on OpenShift clusters #251

### DIFF
--- a/runHarness/ComputeResources/KubernetesCluster.pm
+++ b/runHarness/ComputeResources/KubernetesCluster.pm
@@ -472,7 +472,7 @@ sub kubernetesDeleteAllForCluster {
 	my $initialRemaining = 0;
 	foreach my $namespace (@namespaceNames) {
 		if ($self->useAvailableNamespaces || $namespace =~ /^$matchStr/) {
-			$cmd = "kubectl get deployment,statefulset,service,configmap,pod -o=jsonpath='{.items[*].metadata.name}' --namespace=$namespace --kubeconfig=$kubeconfigFile $contextString";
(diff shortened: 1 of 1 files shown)


ninadingale (approved):

RebeccaYo:
Hi @bhoflich , a couple of questions.
I noticed kristiasty's comment that "I found in the [KubernetesCluster.pm](https://github.com/vmware/weathervane/blob/master/runHarness/ComputeResources/KubernetesCluster.pm) source code there are several calls to kubectl with the --selector=app=auction argument, but for some reason this is missing in the kubernetesDeleteAllForCluster subroutine." I didn't find to be accurate. 
The presence of the --selector=app=auction argument was used only in the collection of the *GetAllPods.txt, *DescribePods.txt etc. files, not in any of the deletion routines. 

kubernetesDeleteAll, kubernetesDeleteAllWithLabel, kubernetesDeleteAllWithLabelAndResourceType could also delete the OpenShift system configmaps as they don't include --selector=app=auction. Are these being used? Do these need the selector added? 

Second, do you think there are any files that should be deleted by kubernetesDeleteAllForCluster that would not get deleted if --selector=app=auction is added?

bhoflich:
> kubernetesDeleteAll, kubernetesDeleteAllWithLabel, kubernetesDeleteAllWithLabelAndResourceType could also delete the OpenShift system configmaps as they don't include --selector=app=auction. Are these being used? Do these need the selector added?

Currently these routines are all being called selectively with parameters passed to them that limit the scope of what is being deleted, and are not deleting the configmaps at the source of the issue to my knowledge. 


> Second, do you think there are any files that should be deleted by kubernetesDeleteAllForCluster that would not get deleted if --selector=app=auction is added?

I did not detect any leftovers in my testing.

